PR #14, opened 2019-03-22
Single alloc gpu

bvillasen:
Instead of allocating GPU memory at the beginning of each time step and freeing the GPU memory at the end of each time step, now if the compiler flag SINGLE_ALLOC_GPU is defined the memory is only allocated at the beginning of the first time step and freed at the end of the simulation.

This was only implemented for VL_3D.

The main changes are:
- Make GPU memory pointers ( dev_conserved, Q_L, Q_R, etc. ) global on VL_3D_cuda.cu file, so that once its values are set at the first time step the pointers remain the same for the entire simulation, instead of redefining them every time step.
- Use a boolean flag "memory_allocated" which is Null the fisrt timestep allowing the GPU memory to be allocated the first time step and setting  memory_allocated to true after the memory is allocated. The next time steps the memory is not allocated since memory_allocated==true
- Create a function Free_Memory_VL_3D() which is called at the end of the simulation and frees the GPU memory allocated for VL_3D
- If the compiler flag SINGLE_ALLOC_GPU is not defined the GPU memory is allocated and freed every time step.


evaneschneider:
Tested new code on a 256^3 sound wave test. Everything seems to be working correctly, but oddly, on a P100 the version _without_ memory allocation every time step is slower (by about 35ms) than the version _with_ memory allocation every step. I'm going to do some more testing of timing before merging this change into master.

bvillasen:
I found that when using 256^3 per gpu the single_alloc change made it slower, but it made it faster when using 128^3 per gpu, if your test was on a single P100, then longer time when single_alloc would be consistent with my tests. Have you tested for 128^3?

evaneschneider:
Yeah, I did a little digging and figured out the issue. When the memory is allocated during the first time step, subsequent time steps are still calling the sub grid splitting function, which looks at the free memory, and concludes that it's full because of the pre-allocated arrays, so then you end up splitting into sub grid blocks, allocating the CPU buffers, etc. etc. This is only an issue for bigger block sizes, because if you're using a 128^3 block there's room for two of them without sub grid splitting. I've done a quick fix on my local version to test it, and when that bug is fixed, the version without array allocation is significantly faster - something like 50% on the P100 with a 256^3 block. I'm going to put together a more robust version of the fix this weekend and then I'll push it to master. I think in that case we don't need to use a flag, it can just be the default because it should always be faster.

bvillasen:
Excellent!

PR #19, opened 2019-08-20
Dual energy fix

bvillasen:
Pressure Calculation:
When using Dual Energy, each time the pressure is computed I apply condition from Bryan+2013 eq 44 to select the internal energy from which the pressure is computed. This was added to the following functions:
- All cuda reconstructions: PLMP, PLMC, PPMP and PPMC
- All cuda Riemann solvers: HLLC, ROE and EXACT
- When adding the term p_div(v) to the advected internal energy update. ( Not applied on the half step update when using VL, We decided not to add the p_div(V) term for the half update, but I didn't change the implementation here )

Internal Energy Synchronization:
Now, there are two steps for the Internal Energy Synchronization:
-First I select wich internal energy to keep by applying the condition from Bryan+2013 eq 45 and write the selected internal energy ONLY to the GasEnergy array. This is because the condition for the Internal Energy selection depends on the neighboring values of the total energy, in order to avoid mixing updated and non-updated values of the total energy, the total energy is updated using the chosen internal energy in a separate kernel.
modified kernels: Select_Internal_Energy_3D   and  Sync_Energies_3D
I only applied this to the 3D versions, 2D and 1D use the original criteria for Dual Energy condition.

Aditional term for Advected Internal Energy:
The Advected Internal Energy is updated using an additional term: p_div(V), originally this term was computed and added on the Update_Conserved_Variables_3D kernel, the problem was that the value of div(V) was computed using the velocities on the conserved array and at the same time the conserved_array was being updated by the same kernel leading to mixing of updated and non-updated values of V for computing div(V). The solution was to add the term p_div(v) in a separate kernel ( Partial_Update_Advected_Internal_Energy_3D ) before the Update_Conserved_3D kernel.
Only implemented for 3D. Another possibility is to use the reconstructed velocities to compute div(V) 

Ghost Cells in PPMP:
For PPMP ( and PPMC ) n_ghost=4, 2 ghost cells on each side. The PPMP kernel is ignoring the first 3 cells and the last 3 cells on each direction, I changed the kernel so that it only ignores the first 2 and last 2 ghost cells, On this implementation FLATTENING is not allowed since it requires 3 ghost cells on each side.

### DIFF
--- a/src/CTU_3D_cuda.cu
+++ b/src/CTU_3D_cuda.cu
@@ -198,7 +198,11 @@ Real CTU_Algorithm_3D_CUDA(Real *host_conserved0, Real *host_conserved1, int nx,
     #endif //HLLC
     CudaCheckError();
     #endif //CTU
-
+    
+    #ifdef DE
+    // Compute the divergence of Vel before updating the conserved array, this solves sincronization issues when adding this term on Update_Conserved_Variables_3D
+    Partial_Update_Advected_Internal_Energy_3D<<<dim1dGrid,dim1dBlock>>>( dev_conserved, Q_Lx, Q_Rx, Q_Ly, Q_Ry, Q_Lz, Q_Rz, nx_s, ny_s, nz_s, n_ghost, dx, dy, dz,  dt, gama, n_fields );
+    #endif
   
     // Step 5: Update the conserved variable array
     Update_Conserved_Variables_3D<<<dim1dGrid,dim1dBlock>>>(dev_conserved, F_x, F_y, F_z, nx_s, ny_s, nz_s, x_off, y_off, z_off, n_ghost, dx, dy, dz, xbound, ybound, zbound, dt, gama, n_fields);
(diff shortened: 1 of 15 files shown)


bvillasen:
OK, you're right,  n_ghost is for each side.
The issue I had was that when using PPMP ignoring 3 ghost cells I got wrong results (surprisingly mass wasn't conserved ) and I fixed it by only ignoring 2 ghost cells ( same as PPMC ).  Perhaps some more tests are needed for that.

evaneschneider:
Hmmm yeah that's weird. Do you get mass conservation with PPMC? If so, I will merge in the pull request as you wrote it and just note that currently shock flattening in PPMP is disabled.

bvillasen:
Yes, mass was conserved with all the other reconstruction methods except for PPMP. 
That simple change fixed it and now I get the correct result.

evaneschneider:
Okay, I'll look into that. As currently written, it seems like all the reconstruction methods require one more ghost cell than is necessary, but it might be because of the centered-difference in the slope reconstruction.

PR #23, opened 2020-03-05
Hipify

ojwg:


evaneschneider:
Just did a first-pass code review. There are still a lot of CudaSafeCall and CudaCheckError functions -do those need to be removed, or does HIP just ignore them?

ojwg:
I think the CudaSafeCall and CudaCheckError functions are OK.  While the names haven't changed the internals have been 'hipified' (for example: inline void __cudaSafeCall( hipError_t err, const char *file, const int line )

reubendb:
For functionality that should be fine. But if Cholla is going to adopt Hip as its portability interface (which is something to decide, I suppose), it'd probably be better to be "CUDA agnostic" as well in the code (for better aesthetic if nothing else). This would be something to do by hand since outside the hipified tool.

evaneschneider:
Got it. So, we'll go ahead and merge this PR so that we can start adding additional functionality, but I agree with Reuben that in the end we'll probably want to be interface agnostic.

dmcdougall:
MPI builds are broken on poplar.  There isn't really a clean way to do this, but I recommend using hipcc to link your code.  Unfortunately, when you do that you lose mpicc/mpicxx's handling of paths to headers and runtime libraries.  The good news is that (at least for OpenMPI) you can get this information by calling `$(shell mpicxx -showme:compile)` in your Makefile.  There's a similar one for link flags too.  Again, not ideal.

It's possible to use a regular compiler (or mpicc/mipcxx) to link instead of hipcc and manually add in paths to HIP's runtime libraries but there are some speedbumps when you do this.  We can go down that rabbit hole if you'd like, but my preliminary exploration I can see that it requires some Cholla code changes but I don't know how invasive it would be.  It might also require some ABI compatibility between various language runtimes, and that gets hairy.

I'll write up a PR that fixes the MPI builds using the strategy in the first paragraph above.

dmcdougall:
Actually, depending on the MPI stack, you can modify the compiler that the mpicc/mpicxx wrappers call.  For OpenMPI you can build it like this:

```
OMPI_CC=/opt/rocm/hcc/bin/clang OMPI_CXX=/opt/rocm/bin/hipcc make
```

And that works on poplar on the CAAR branch.  This requires some ABI compatibility, I don't know all of the details yet.

PR #29, opened 2020-06-25
Cholla pm paris hip

bvillasen:
Added SOR Poisson Solver

Changed Makefile To Include flags for Particles and Cosmology

Added "System"  to make file to simplify the compiling on different systems.

To make in my local system ( lux ) I just do: sh make_lux_sor.sh


bvillasen:
we'll leave this merge for latter once the new gpu calls have been hipified

PR #50, opened 2020-09-10
Added Paris support for nonzero lower domain bounds and nonunit domain lengths.

twhite-cray:


twhite-cray:
I tried to create a new pull request today, but somehow it all got added to this one. The additional stuff fixes the build on Poplar after the upgrade, adds support for MI100, and moves all the build and run scripts in the root directory to subdirectories.

PR #55, opened 2020-12-01
Numdiff and Makefile mods

reubendb:
This PR contains python script `numdiff.py` to compare files numerically for regression testing and other minor modifications in Makefile and build-related scripts for Poplar compilers.

### DIFF
--- a/builds/scripts/amake-paris-pfft.sh
+++ b/builds/scripts/amake-paris-pfft.sh
@@ -1,4 +1,5 @@
 #!/bin/bash
+module restore PrgEnv-cray
 module load pfft
 module load hdf5
 module load gcc/8.1.0
--- a/builds/scripts/amake-paris-pfft.sh
+++ b/builds/scripts/amake-paris-pfft.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-module restore PrgEnv-cray
+module use /home/users/twhite/share/modulefiles
 module load pfft
 module load hdf5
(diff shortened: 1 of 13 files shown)


builds/scripts/amake-paris-pfft.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+module restore PrgEnv-cray
```
twhite-cray, line 2: Did you try this change? The `pfft` module already does a `module restore PrgEnv-cray`. And I think doing `module restore PrgEnv-cray` first will wipe out the `module use` that allows you to see the `pfft` module.

reubendb: For some reason I can't see `pfft` module either way, so with / without the `module restore PrgEnv-cray` this script fails to build. Is there a missing `module use ...` that is supposed to be there?

twhite-cray: Yeah, sorry, it's in my default environment.
```
module use /home/users/twhite/share/modulefiles
```

reubendb: OK, thanks. I think e2533b0 fixes this. The scripts now build correctly for me.

reubendb: I am not sure why `module restore` wipes modulepath added by `module use`; I don't think I'd agree with that behavior but I guess that's a side point ...

PR #76, opened 2021-08-11
Remove dependence on offload

alwinm:
- Replaces OpenMP offload with CUDA
- new kernels are written in cuda_pack_buffers.cu
- Changed loading of ghost cells with real cell values in boundary_conditions.cpp Grid3D::Set_Boundaries
- Changed loading/unloading of MPI buffers in mpi_boundaries.cpp Load_Hydro_DeviceBuffer and Unload_Hydro_DeviceBuffer functions

More testing is required. Initial tests on 2D Gresho and 3D Spherical_Overpressure_3D. 

Preliminary timing tests indicate that changes to both files result in a faster boundary update step. 

Some cleanup of CUDA functions, old openMP pragmas may be called for


twhite-cray (commented):
Changing the OpenMP offload to Cuda/Hip could be simpler by using the existing `gpuFor` templates in `gpu.hpp`. You wouldn't need any new files.
Here's an example change. Code like this:
```
    #pragma omp target teams distribute parallel for \
             private ( idx, gidx ) \
             firstprivate ( offset ) \
             is_device_ptr ( send_buffer_x0, c_head )
     for (i=0;i<H.n_ghost;i++) {
       idx = (i+H.n_ghost);
       gidx = i;
       for (ii=0; ii<H.n_fields; ii++) {
         *(send_buffer_x0 + gidx + ii*offset) = c_head[idx + ii*H.n_cells];
       }
     }
   }
```
... would become code like this:
```
     const int ng = H.n_ghost; // need local copies of class members for lambda capture
     const int nf = H.n_fields;
     const int nc = H.n_cells;
     gpuFor(
       nf,ng,
       GPU_LAMBDA(const int ii, const int i) {
         send_buffer_x0[i+ii*offset] = c_head[i+ng+ii*nc];
       });
```
You would need to change the name `mpi_boundaries.cpp` to `mpi_boundaries.cu` so it's compiled with the Cuda and Hip compilers. All the changes would be local to that file.
You can `grep` the existing CAAR branch for `gpuFor` to see more examples of its use. And I'm happy to answer questions about it here or on Slack.

alwinm:
I'll preface my comment by disclosing I'm not a software development expert or a CUDA expert. In both, I am a novice!

I'll explain some of my rationale for some of the choices I made. 

My personal opinion is that changing a .cpp file to a .cu file and making all the changes locally is a larger change. Whereas adding a new file and containing most of the new code inside it feels somewhat more modular. I tried to change the existing files as little as possible, mainly deleting things only if Evan's told me to. As a result, I've even left in some code that my changes should mark for deletion. 

In mpi_boundaries.cpp there are 3 for loops in "x", 2 for loops in "y" and 1 for loop in "z" nested to varying degrees based on dimension, which have all been abstracted into a single function, reducing duplicate code. The function declarations are ugly and long, but potentially easier to check, since only a few real variables change between all these loops. 

In boundary_conditions.cpp I moved more of the load onto the GPU, which required making copies of some Grid3D functions, as I'm not sure how to simply give those functions to a GPU to use. In general, most of the extra lines I'm adding are contained within cuda_pack_buffers.cu due to those copies. I admit that this isn't great, but it does make the boundary step faster. For this more complicated section, I don't have any clue how I would do it with gpuFor, because of my lack of familiarity with CUDA and calling host vs device functions. 

I did see the gpuFor template when I was looking through the code for CUDA examples to learn from. It looks very useful but it's out of my expertise / comfort level to try to use, for now. As I get more comfortable with CUDA I'll hope to bug you about it--it feels very much like a convenient "map" command.

PR #78, opened 2021-08-18
Caar2

ojwg:
changes needed to run particles-only disk.  Preprocessor flags are collected in builds/make.type.disk, so to build invoke `make TYPE=disk`


evaneschneider (commented):
This all looks good, my only comment is that I think you should rename parameter_file.txt something more meaningful. But I'll merge the PR, regardless.

PR #83, opened 2021-08-31
Clean boundaries

alwinm:
- Fixed transmissive boundaries in boundary_conditions.cpp and cuda_pack_buffers.cu to avoid floating point errors by performing fewer FLOPs (output matches much better between HYDRO_GPU on/off). This is the most significant change in terms of output differences from previous versions of the code, but should be more accurate. (boundary_conditions.cpp and cuda_pack_buffers.cu)
- After more testing, I realized the CUDA version of the code requires thread synchronization as well, and previous tests somehow got lucky. (cuda_pack_buffers.cu)
- Cleanup of a lot of trash code I generated when making no_offload changes (mpi_boundaries.cpp)
- Enable HYDRO_GPU to be turned off while MPI_GPU is turned on (for hydro buffers, it will not use cuda-aware MPI) (mpi_boundaries.cpp)
- Refactor of mpi_boundary CPU buffer packing/unpacking to look more similar to the GPU buffer packing/unpacking to make reading code / changing code / debugging code easier (if someone wants to, maybe they can insert an openmp pragma to make CPU buffer packing/unpacking faster. Since the loops are abstracted, the pragma only needs to be placed in a few locations). (mpi_boundaries.cpp)
- All changes tested with 2D and 3D multi-GPU on summit/spock

### DIFF
--- a/src/boundary_conditions.cpp
+++ b/src/boundary_conditions.cpp
@@ -389,6 +389,26 @@ void Grid3D::Set_Hydro_Boundaries_CPU
     //for outflow boundaries, set momentum to restrict inflow
     
     if (flags[dir] == 3) {
+      int momdex = iaBoundary[iB] + (dir/2+1)*H.n_cells;
+      // (X) Dir 0,1 -> Mom 1 -> c_density[gidx+1*n_cells]
+      // (Y) Dir 2,3 -> Mom 2 -> c_density[gidx+2*n_cells]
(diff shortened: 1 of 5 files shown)


alwinm:
main_vs_new
2D sod test comparing main branch with new treatment of transmissive boundaries
Mean, standard deviation are taken across the Y-axis. I believe lower standard deviation across Y-axis means the 2-D problem is behaving more closely to 1-D. New treatment of transmissive boundaries preserves momentum_Y = 0 for longer time.

PR #84, opened 2021-09-01
Source Files Reorganization

bcaddy:
- Moved all the source files in `src` to appropriate subdirectories. This included making several new directories. The only remaining file in `src` is `main.cpp`
- Deleted the `.gitignore` and `src.udb` files within `src`
- Deleted spare copy of `cloud_coolingcurve.text` that was in `src`. The primary copy is in `src/cooling`


src/.gitignore
```diff
@@ -1 +0,0 @@
-*.o
```
twhite-cray, line 1: This file hides all the `.o` files from a build when running `git` commands. Without it, won't all the `.o` files clog up `git status` and `git commit` commands? We do not want to accidentally commit `.o` files to this repo. Why was the file removed?

twhite-cray: Oops. Looks like there's a top-level `.gitignore` that hides `.o`, and this one is redundant. Right?

bcaddy:
That's correct. The top level .gitignore​ takes care of .o files so I removed this extranous one.

________________________________

bcaddy (commented):
That's correct.

PR #87, opened 2021-09-16
Command line override args

alwinm:
This minor patch allows command line overrides, which is a quality-of-life convenience when testing and can reduce the # of input files needed to run a variety of tests. Use examples:

Changing the output directory 
`./bin/cholla tests/3D/constant.txt outdir=dir2/`

Changing the resolution
`./bin/cholla tests/3D/constant.txt nx=100 ny=100 nz=100`

Changing tout
`./bin/cholla tests/3D/constant.txt tout=1000.0 outstep=1000.0`

Changing parameters
`./bin/cholla tests/3D/constant.txt rho=1e2`

Everything combined
`./bin/cholla tests/3D/constant.txt outdir=dir2/ nx=100 ny=100 nz=100 tout=1000.0 rho=1e2`

Adding new params to the codebase should not be any different from before, as there is only one copy of the param parser, which is first used on the input file and then used on command line arguments, which gives command line arguments the power to override.

### DIFF
--- a/src/global/global.cpp
+++ b/src/global/global.cpp
@@ -170,6 +170,7 @@ parms->scale_outputs_file[0] = '\0';
   /* Close file */
   fclose (fp);
 
+  // Parse overriding args from command line
   for (int i = 0; i < argc; ++i) {
     char name[MAXLEN], value[MAXLEN];
     s = strtok (argv[i], "=");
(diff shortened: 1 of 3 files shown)


evaneschneider:
Great idea.

evaneschneider (approved):

PR #99, opened 2021-11-08
MPI supported unique particles IDs

bvillasen:
- Now when using multiple MPI ranks, the particles IDs are offset such that global unique IDs are generated. Applied to Uniform Sphere and Disk initialization routines.

- Fixed bug when initializing the particles IDs for the Disk problem

- Updated FOM makefile to the latest FOM test I ran in July 2021.


bcaddy:
It looks like the failing builds are probably due to a bug in GCC 10.3 and possibly some CUDA incompatibility. I've, hopefully, got a fix in PR #100 by just switching to GCC 9 in the builds. Once that's accepted then rebasing your branch off of the root CAAR branch should fix it. If the checks don't automatically rerun after your rebase then [go here](https://github.com/cholla-hydro/cholla/pull/99/checks) and select "re-run all jobs"

src/mpi/mpi_routines.cpp
```diff
@@ -788,6 +788,27 @@ Real ReducePartIntSum(part_int_t x)
   y = (part_int_t) out ;
   return y;
 }
+
+
+// Count the particles in the MPI ranks lower that this rank (procID) to get a global offset for the local IDs.
```
twhite-cray, line 793: Minor typo in comment. "lower tha**t** this rank"

PR #102, opened 2021-11-09
Revert "Revert "MPI supported unique particles IDs""

bcaddy:
Reverts cholla-hydro/cholla#101

I think the issues that caused PR #99 to fail the first time have been resolved by PR #100. Hopefully this will pass


twhite-cray:
I had comments on PR #99 that apply to this request as well.

bcaddy:
> I had comments on PR #99 that apply to this request as well.

Good catch. To double check, is it supposed to be "lower than this rank"? I've got a PR coming up that I can add this fix to

bvillasen:
Yes, “lower than this rank” is what it should say.

bcaddy:
Alright, I'll fix that and add it to my PR

PR #116, opened 2022-02-04
GPU solver for H and He chemistry, cooling and photoheating

bvillasen:
Included a solver for the non-equilibrium ionization states of hydrogen and helium that also computes the radiative cooling and photoheating.

The ionization states of H+He are tracked as scalar fields and advected with the hydro solver. Each time step the ionization states are updated based on the ionization and recombination rates as well as photoionization, additionally, the cooling rates are computed and the internal energy is updated.

Cooling processes:
- Collisional excitation
- Collisional ionization
- Recombination
- Bremsstrahlung
- Compton cooling from the CMB
- Photoheating from the UV background (time-dependent photoheating and photoionization rates are passed by the user)

The solver closely follows the implementation from Grackle. A comparison from simulations using Grackle and the new GPU solver can be found here: 
https://bvillasen.github.io/blog/astro/2022/01/28/cholla_chemistry_new.html

The solver can be used with or without HYDRO_GPU. 

This PR also includes updates made to the Analysis module for cosmological simulations.


bvillasen:
There is a bud when COSMOLOGY is turned off. Fixing and resubmitting soon

PR #118, opened 2022-02-18
Hydro without MPI

alwinm:
This PR confirms that I was wrong in the CAAR meeting -- currently the CAAR branch will not compile without MPI. 

I added two protections (and a soon-to-be-deprecated include to subgrid routines), and now (CAAR branch + hydro only) will work without MPI (MPI_CHOLLA off, MPI_GPU off, and compiler set to g++ instead of mpicc(x)). 

One outstanding but scientifically unimportant issue is that the "domain" attribute of HDF5 output differs for irrelevant dimensions (Y,Z for 1D, Z for 2D, no difference for 3D) between MPI and non-MPI. This is based on how the variables H.ydglobal and H.zdglobal are set.

### DIFF
--- a/src/io/io.cpp
+++ b/src/io/io.cpp
@@ -136,7 +136,9 @@ void WriteData(Grid3D &G, struct parameters P, int nfile)
   G.H.Output_Now = false;
   #endif
 
+  #ifdef MPI_CHOLLA
   MPI_Barrier(world);
+  #endif
 }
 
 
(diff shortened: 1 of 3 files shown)


bcaddy:
I found the same issue with the weekly builds, I'll let you know when I've got it sorted

bcaddy:
@alwinm, This should be fixed with PR #119. Once that gets merged in I'll rerun the checks for this PR

PR #123, opened 2022-02-23
CAAR timing refactor

alwinm:
Refactor Timing to make it a little easier to add/remove timers (and reduces lines of code). 

Timers also now come with names to make it easy to see what is being timed. 

And Timers can be nested (if you wanted to, you could time MPI, MPI_X, MPI_Y, MPI_Z. Or you can time HYDRO, HYDRO_INTEGRATOR, HYDRO_CHEMISTRY) since each Timer keeps track of its own start_time. 

I need some input from @bvillasen -- CHEMISTRY_GPU adds a more complicated timer and I tried to capture the spirit of it but I would like to confirm, and I believe Bruno was the last person to refactor the Timers. 

I am also thinking that the Total Timestep should be its own Timer and that will allow me to remove the step where we calculate the Total Timestep Time by summing over the others. Since Timers can be nested, the summing approach is no longer necessary.


bvillasen:
@alwinm, I like these changes... when I originally designed the timing scheme I didn't anticipate the large number of timers that we had now and I agree that it had become quite cumbersome to add new timers. 

As you said, there is some separation of the Hydro and Chemistry timers, seems to me that your implementation is correct but I'll have to run it to confirm.

I support the idea of adding a timer for the global time step instead of summing the individual components.

... this looks good, no complaints from me :)

evaneschneider:
@bvillasen Do you want me to wait to merge this PR until you get a chance to test it? Or are you okay with merging it in and you can alert us if you find any bugs?

bvillasen:
@evaneschneider, I'm ok with merging now and I'll report back when I get some time to test

alwinm:
@evaneschneider I'm going to touch up the PR a little bit by making a timer for the total time step instead of summing the individual components. 

@bvillasen After I replace the total time step, would the subtraction (hydro - chemistry) step still be necessary? I guess I am asking if the subtraction was just to maintain the total, or if the subtraction was also to isolate the hydro step.

evaneschneider:
@alwinm Happy to wait until you have it in whatever form you like :)

bvillasen:
@alwinm, subtracting the Chemistry timer from the Hydro is necessary because the Chemistry_Update happens inside the Hydro_Update, that way you don't include the in the hydro time the time spent on chemistry. So doing the timing of the global step will still require subtracting the chemistry time from hydro, does that make sense?

PR #124, opened 2022-02-25
add constant and sound 3D test

alwinm:
Also demonstrates an example of running analytic tests in hydro_system_tests.cpp,
adds a few helper functions,
and makes _loadTestFieldData public.

### DIFF
--- a/src/utils/testing_utilities.cpp
+++ b/src/utils/testing_utilities.cpp
@@ -10,9 +10,14 @@
 #include <limits>
 #include <cmath>
 #include <cstring>
+#include <iostream>
+
+// External Includes
+#include <gtest/gtest.h>    // Include GoogleTest and related libraries/headers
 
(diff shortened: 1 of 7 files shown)


src/utils/testing_utilities.cpp
```diff
@@ -64,4 +69,66 @@ namespace testingUtilities
         return ulpsDiff <= ulpsEpsilon;
     }
     // =========================================================================
-}
\ No newline at end of file
+
+  void wrapperEqual(int i, int j, int k, std::string dataSetName,
```
bcaddy, line 73: I've got a more generic version of this function already implemented that I will be submitting a PR for. See [`checkResults`](https://github.com/bcaddy/cholla/blob/790f7323e0882490f6f18227cef1e70a304916f7/src/utils/testing_utilities.cpp#L73)

PR #129, opened 2022-03-02
New test for hydro 3D time step kernel

evaneschneider:


src/hydro/hydro_cuda_tests.cu
```diff
@@ -0,0 +1,90 @@
...
+#include "../global/global.h"
+#include "../global/global_cuda.h"
+#include "../utils/gpu.hpp"
+#include "../utils/testing_utilities.h"
+#include "../hydro/hydro_cuda.h"   // Include code to test
+
+#if defined(CUDA)
+
+
+TEST(tHYDROCalcDt3D, CorrectInputExpectCorrectOutput)
```
bcaddy, line 27: Should this run for other make types besides only hydro?

src/hydro/hydro_cuda_tests.cu
```diff
@@ -0,0 +1,90 @@
...
+  double testData = testDt[0];
+  double absoluteDiff;
+  int64_t ulpsDiff;
+  bool areEqual; 
+  areEqual = testingUtilities::nearlyEqualDbl(fiducialDt, testData, absoluteDiff, ulpsDiff);	
+  EXPECT_TRUE(areEqual)
+    << "The fiducial value is:       " << fiducialDt << std::endl
+    << "The test value is:           " << testData     << std::endl
+    << "The absolute difference is:  " << absoluteDiff << std::endl
+    << "The ULP difference is:       " << ulpsDiff     << std::endl;
```
bcaddy, line 85: The `checkResults` function in [`testing_utilities.cpp`](https://github.com/cholla-hydro/cholla/blob/729856eeee1ef2b2a9e0d18a4fa1f0268adacce5/src/utils/testing_utilities.cpp#L73) does all this if you want to avoid boilerplate. That function is nice when you're writing a lot of tests and don't want hundreds of lines of the same boilerplate code repeated over and over again

bcaddy (commented):
Looks good. I made a couple of comments but nothing really looks like an issue

src/hydro/hydro_cuda_tests.cu
evaneschneider: It should run for any make type that includes hydro, so the way the current builds are defined, that would include gravity and particles, but it doesn't have to, in principle (one could build with gravity and particles but no hydro, and this function would not get called).

src/hydro/hydro_cuda_tests.cu
evaneschneider: Do you have an example using it? I tried to follow in hllc_cuda_tests but it looks like there a different version of the function is being defined.

src/hydro/hydro_cuda_tests.cu
bcaddy: Ok, as is it will only run with the make type "hydro". If it should run with other make types then you should add them to the suite name

src/hydro/hydro_cuda_tests.cu
bcaddy: [here](https://github.com/bcaddy/cholla/blob/790f7323e0882490f6f18227cef1e70a304916f7/src/utils/mhd_utilities_tests.cpp#L75). I don't think there's any examples on CAAR but I've used it a lot on my MHD branch

The duplicated version of it I should probably remove or rename

src/hydro/hydro_cuda_tests.cu
evaneschneider: I think because they could be redefined without the hydro dependence, it makes more sense to confine it to hydro.

src/hydro/hydro_cuda_tests.cu
evaneschneider: Okay. I still don't really understand how to call it (does it need a string argument?), so I think I'll leave my test as-is for now, and maybe later we can add it in as an example.

PR #133, opened 2022-03-04
Move all timestep calculations to GPU

bcaddy:
Passes All Tests

- Switched the first time step calculation to be done on GPU rather
  than CPU
- Make the `host` and `device` `dti_array`'s local to `Calc_dt_GPU`
- Removed last reference to `HYDRO_GPU` flag in cooling make file
- Added comments to main to clarify what `Grid3D::set_dt` does
- Deleted CPU time step functions
  - `calc_dti_cpu`
  - `calc_dti_cpu_1D`
  - `calc_dti_cpu_2D`
  - `calc_dti_cpu_3D`
  - `calc_dti_CPU_3D_function`

### DIFF
--- a/builds/make.type.cooling
+++ b/builds/make.type.cooling
@@ -1,14 +1,13 @@
 #-- Default hydro + cooling_gpu
 
-#-- separated output flag so that it can be overriden in target-specific 
+#-- separated output flag so that it can be overriden in target-specific
 #   for make check
 OUTPUT    ?=  -DOUTPUT -DHDF5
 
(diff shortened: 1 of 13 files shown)


evaneschneider:
I looked it over, and it's mostly fine. This line "if (block_tot > 1) CudaSafeCall( cudaFreeHost(buffer) );" should not be in VL_3D_cuda (or any of the other integrators) any more.

bcaddy:
Fixed it, I probably just missed that line when rebasing

Issue #98, opened 2021-11-03
Duplicated Particle IDs


bcaddy:
I found a lot of duplication of particle IDs when running the disk test in [examples/scripts/parameter_file.txt](https://github.com/cholla-hydro/cholla/blob/ce6646e9f04c19d4851bc6172afcfae93c495543/examples/scripts/parameter_file.txt). I've attached an example of the `particle_IDs` dataset in the HDF5 file from running on CRC, the same issue does persist on Summit and Spock though exactly which IDs are duplicated does not appear to be consistent

[particle_ids.txt.zip](https://github.com/cholla-hydro/cholla/files/7469450/particle_ids.txt.zip)

bcaddy:
Update: The IDs and duplication therein vary from run to run even when the random seed is set to be fixed

bcaddy:
Per a discussion with Bruno and Orlando: The issue with duplicated particle IDs per rank is solved with converting `Real` to `part_int_t` on line 575 of `particles_3d.cpp` and that change will be in future PR from Bruno

bcaddy:
I think this issue has be resolved so I'm closing it

PR #134, opened 2022-03-07
CAAR Testing Data Submodule + Testing Fixes

bcaddy:
# Add cholla-tests-data as a submodule

After we've added many system tests the total size of all the fiducial data files could get fairly large. To avoid cloning all of these, and their histories, on every clone of Cholla the data files used for testing are now stored in the cholla-testing-data repository which is connected to the Cholla repo as a submodule. Please see the [cholla-testing-data](https://github.com/cholla-hydro/cholla-tests-data) README for details on how to clone it. The cholla-testing-data repo also stores all HDF5 files using Git LFS to reduce the size of the repo and the size of each clone. Please see that repos README for details on Git LFS, Atlassian also has excellent documentation.

- Removed fiducial HDF5 files from main cholla repo
- Update fiducial data file path for new submodule

# Fix systemTestRunner requiring fiducial HDF5 file

There was an implicit assumption in `systemTest::SystemTestRunner::runTest()`
and `systemTest::SystemTestRunner::_loadTestParticleData()` that a
fiducial HDF5 file existed. I believe I've removed that assumption.

# Fix particles spherical collapse test

The existing particles test has no fiducial data due to issue #98. Now that the issue has been resolved I added the fiducial data file and the test should pass.

# Automated GHA builds

Set builds to automatically run weekly, I've been doing this manually but since it looks like we'll be keeping at least the HIP builds long term I'm setting it to be automated

# Other
- Add particle_ids to default particle build, currently required for testing. If this requirement becomes an issue please let me know
- Add assertions to check that Cholla was launched properly and to provide clear output if/when Cholla fails to launch


evaneschneider:
Does this PR mean that the current documentation at the beginning of the testing wiki is out of date? That is, now if a user tries to build and run the tests as currently specified, system tests will not work (until the submodule is cloned)? Or have those functions been changed to not run system tests?

bcaddy:
Yes, the current documentation will be out of date once this PR is merged. I have updated documentation that I’ll post once the PR is in. I can send it your way first if you’d prefer

bcaddy:
Documentation is up. Wiki commit 3993af4 and 68a1238

PR #135, opened 2022-03-08
Caar

evaneschneider:
Moved the memory allocations for host and device dti arrays to grid3D.cpp, so they are single allocated. These are used for the GPU time step reduction in Calc_dt_GPU. Also, the cooling dti calculation now re-uses the same arrays (since both are only used in the kernel).
Added definition of global variable 'ngrid' - the number of GPU thread blocks - to grid3D.cpp. Removed duplicate definitions in hydro integrators.
Removed duplicate hydro conserved variable array on the host (in grid3D.cpp).


bcaddy (approved):
Everything looks fine to me. I'll have to change some stuff with how magnetic fields are allocated but it should just be a couple lines of converting `buffer0` to `C.host`.

Does this mean the `buffer0`/`buffer1` stuff is totally gone and I don't have to worry about it for magnetic fields?

evaneschneider:
That's right, no more buffers. Thanks!

PR #139, opened 2022-03-24
fix particle id type mismatch and discarded boundary particles problem

ojwg:
Fix from Trey to prevent particles on MPI boundaries from being discarded and a second fix to keep particle IDs as int

### DIFF
--- a/src/particles/particles_3D.cpp
+++ b/src/particles/particles_3D.cpp
@@ -675,26 +675,9 @@ void Particles_3D::Initialize_Disk_Stellar_Clusters(struct parameters *P) {
       #endif//PARTICLES_CPU
   }
 
-
-
   #ifdef PARTICLES_CPU
   n_local = pos_x.size();
   #endif
-
-  #if defined(PARTICLE_IDS) && defined(MPI_CHOLLA)
-  // Get global IDs: Offset the local IDs to get unique global IDs across the MPI ranks
-  chprintf( " Computing Global Particles IDs offset \n" );
-  part_int_t global_id_offset;
-  global_id_offset = Get_Particles_IDs_Global_MPI_Offset( n_local );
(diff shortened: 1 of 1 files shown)


bvillasen:
seems to me like the change proposed will cause only half of the MPI tasks to have particles when running with 8 MPI tasks. Only the MPI tasks for which their domain is in the z>=0 side of the box will have particles, the other 4 MPI tasks for which G.zMax=0 will discard all their particles in the initialization loop, this would reduce the number of total particles in the simulation by 1/2. 

Also, it seems like the number of particles per MPI task is constant (N), this means that a run with 2 tasks would have half the number of particles compared to the same simulation but run with 4 MPI tasks, is this intended behavior?

twhite-cray:
I think each task owns domain [min,max). The change to z >= 0 gives behavior consistent with having an odd number of MPI tasks in the Z dimension, where all the particles are at a single level in the Z. Whereas each task creates N particles, it only keeps the particles that reside in its domain. Because the domains meet precisely but do not overlap (with the above code change), I think that the code should create a global total of roughly N particles (exactly N if all the tasks use the same seed).

This redundant computation could get quite expensive at scale.

bvillasen:
Trey, you're right about the total number N being constant regarless of the number of tasks, my bad.

Yes, the domains were designed to be [min, max). One slighly more complicated way to solve the issue of only the top z>=0 tasks having particles when runing 8 tasks would be to add a small z displacemnt epsilon (maybe 1e-8 or smaller) such that half the particles get z=epsilon and the other half get z=-epsilon, and the sign is assigned randomly, that way all 8 tasks would have  particles, and if epsilon is small the particles shouldn't change their dynamics. 

Anyway, this is not that important as the disk will eventually have some thickness and this wouldn't be an issue anymore

PR #142, opened 2022-03-28
Caar domain

alwinm:
- Improved Get_Position so that multi-GPU and single-GPU match perfectly for Rayleigh_Taylor problem (and generally, anything that relies on Get_Position returning the exact same position for the same cell whether it's single-GPU or multi-GPU). 
- Remove Set_Parallel_Domain, move functionality into src/grid/initial_conditions.cpp @@ Grid3D::Set_Domain_Properties
- Remove domlen_x/y/z
- Particles x/y/zMax inherits from Grav
- Grav x/y/zMax inherits from Grid
- This way, all domain xMin and xMax are being controlled by Grid3D::Set_Domain_Properties. 

The key motivation for this rework is that the distributive law does not hold for floating point operations. That is:

```a + (b1 + b2) * c != a + b1*c + b2*c. ```

[Here is a demo!](https://tio.run/##Sy4o0E1PTv7/PzOvRCE3MTNPQ7OaS4FLISW/NCknVSGlQsFWwUDPzNgaLlRcklhUAhTVNTTQMwAK6@sjlCcCxSHy2gqmWkDNEAqhOQlJgaEBRKootaS0KE9BI9HWNknTmqv2/38A)

Therefore, terms of the form:
```local_bound + local_i*dx = (global_bound + (global_i-local_i)*dx) + local_i*dx. != global_bound + global_i*dx. ```
You can see that on the left side of this equation is the multi-GPU calculation for Get_Position, and on the right side is the single-GPU calculation. 

Furthermore, this affects xMin and xMax. If block A is next to block B so that you'd want xMax_A == xMin_B, you should calculate both using globally agreed-upon inputs. Hence, we should replace anything of the form xMax = xlocal+domlen with 
xMax = xglobal + global_i*dx. 

I believe this increased consistency may also help with particle behavior as they cross boundaries. Whenever xMax_A != xMin_B, particles may be stranded or duplicated, leading to potential issues. If possible I think @ojwg should take a look and see if this patch helps with particle behavior. 

I'm leaving this as a draft for now until any stakeholders in Grav/Particles or users of domlen have a chance to take a look. 

- A minor change: whenever the cholla binary is called without arguments, exit immediately (as it always has), but also print git hash and compile options.

### DIFF
--- a/src/gravity/grav3D.cpp
+++ b/src/gravity/grav3D.cpp
@@ -17,7 +17,7 @@
 
 Grav3D::Grav3D( void ){}
 
-void Grav3D::Initialize( Real x_min, Real y_min, Real z_min, Real Lx, Real Ly, Real Lz, int nx, int ny, int nz, int nx_real, int ny_real, int nz_real, Real dx_real, Real dy_real, Real dz_real, int n_ghost_pot_offset, struct parameters *P )
+void Grav3D::Initialize( Real x_min, Real y_min, Real z_min, Real x_max, Real y_max, Real z_max, Real Lx, Real Ly, Real Lz, int nx, int ny, int nz, int nx_real, int ny_real, int nz_real, Real dx_real, Real dy_real, Real dz_real, int n_ghost_pot_offset, struct parameters *P )
 {
 
   //Set Box Size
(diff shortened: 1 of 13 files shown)


ojwg (approved):
these changes look good to me

ojwg (approved):
looks good

PR #145, opened 2022-04-07
Fix HDF5 Write Errors at Larger Scale

reubendb:
This PR fixes errors with HDF5 write that seems to occur at larger scale on ORNL test system. It also includes:
- Typo fix for STDOUT message
- Minor build scripts update


evaneschneider:
If the H5open() and H5close() calls are needed to fix this bug, do we need them in the other output routines, too? (I.e. slices and projections, in addition to the full grid)

reubendb:
Good point; Likely good idea to add these calls there too. Let me do that.

PR #149, opened 2022-05-02
Particles_CPU with Everything else GPU (fixed testing)

bvillasen:
This is the first part of the new version of the particles solver where by default everything is on the GPU and when the global memory is full the excess particles are offloaded to the CPU.

Here the CPU version of particles is compatible with the GPU version of everything else (Hydro, gravity and mpi).

Also, other small changes include separating the Average_Slow_Cells from the calculation of dt as it didn't work with the new version of the reduction to compute the minimum delta_t. Now there is an independent kernel to evaluate if a cells dt is smaller than a minimum and if so then the conserved properties of the cells are replaced with the averaged properties of the neighbours, this is to avoid tiny delta_t values from a single cell.

The pull request with the offloading of excess GPU particles to the CPU will be coming soon.


evaneschneider:
I checked it -- everything passes.

PR #152, opened 2022-05-12
Light hotfix for Dual Energy in initial conditions

alwinm:
Removes vestigial cuda_pack_buffers.h file (somehow came back after being renamed to cuda_boundaries)

Hotfixes KH_res_ind initial condition (which results in NaNs in 3-D) by adding Dual Energy. 

Adds Dual Energy to more of the initial conditions to avoid running into similar problems in the future, light refactoring of assigning energy by taking it out of if-statements when applicable. 

Someone more familiar with Dual Energy should take a brief look, I just assumed GasEnergy = P/(1-gamma).


evaneschneider:
Looks good to me!

PR #154, opened 2022-05-23
Paris Periodic using FFT Filter

trey-ornl:
Changed the Paris periodic Poisson solver from a monolithic object to a generalized FFT filter with a lambda function to perform the Poisson solve. The hope is for the generalized filter class, Henry, to be re-used to create initial conditions at runtime.

### DIFF
--- a/src/system_tests/gravity_system_tests.cpp
+++ b/src/system_tests/gravity_system_tests.cpp
@@ -1,0 +1,30 @@
+/*!
+ * \file gravity_system_tests.cpp
+ * \author Evan Schneider (eschneider@pitt.edu)
+ * \brief Contains all the system tests for the GRAVITY build type
+ *
+ */
+
+// External Libraries and Headers
+#include <gtest/gtest.h>
+
+// Local includes
+#include "../system_tests/system_tester.h"
+
+// =============================================================================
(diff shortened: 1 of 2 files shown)


src/gravity/paris/HenryPeriodic.hpp
```diff
@@ -1,118 +1,39 @@
-#ifdef PARIS
```
twhite-cray, line 1: I don't know why Git made this a file change. This is a separate file, and PoissonPeriodic3x1DBlockedGPU.cu is deleted.

src/gravity/paris/ParisPeriodic.cu
```diff
@@ -0,0 +1,74 @@
...
+        const double i2 = sqr(double(min(i,ni-i))*ddi);
+        const double j2 = sqr(double(min(j,nj-j))*ddj);
+        const double k2 = sqr(double(k)*ddk);
+#endif
+        const double d = -1.0/(i2+j2+k2);
+        return cufftDoubleComplex{d*b.x,d*b.y};
+      } else {
+        return cufftDoubleComplex{0.0,0.0};
+      }
+    });
```
twhite-cray, line 71: Here's an example of specializing the Henry class, a generalized FFT filter, for a specific filter, in this case, a Poisson solve.

PR #156, opened 2022-05-24
Analytic Cooling Test + Fix calc_dt bug

alwinm:
Many technical details below. In my opinion this PR is ready to merge. 

**Analytic Cooling Test**

I included an Analytic Cooling test. Currently it only serves as a regression test, since I'm unable to get precisely the same values (but similar to several decimal places) when using Python to do the same calculation. In order to test CIE_COOL, I chose 3 temperatures to test the 3 cooling ranges.

 **Patch calc_dt to ensure Cholla consistency**

A bug in Nvidia's atomic function mysteriously causes Cholla to produce different output when running with the same binary and inputs. In order to ensure hydro Cholla is reproducible and deterministic,  I fixed up calc_dt in various ways to patch this bug. This unfortunately requires bringing back dev_dti_array. 

**Notes on changes to outputs since PR 146**

Hydro example problems run with this PR will generally match exactly with runs using https://github.com/cholla-hydro/cholla/pull/146 . 

However, there are expected changes in gasEnergy fields printed at t=0 due to updates to initial_conditions.cpp. 

These changes generally do not cause changes at later times, but runs using the "simple" integrator experience gasEnergy changes for the Kelvin-Helmholtz and Sound Wave example problems at later times. 

There are real changes in density, momentum, and energy for the Blast1D problem. This is to be expected, because GasEnergy was not being initialized correctly prior to the dual energy hotfix of https://github.com/cholla-hydro/cholla/pull/152


evaneschneider:
I merged in Bob's changes to load googletest, but that seems to have broken some of the HIP builds.

bcaddy:
> I merged in Bob's changes to load googletest, but that seems to have broken some of the HIP builds.

Looks like it was just an issue with the GitHub Actions servers not being able to find the ROCm servers for ROCm installations. Rerunning the checks fixed it

evaneschneider:
Super, I'll merge!

PR #160, opened 2022-06-17
Deprecate CUFFT, PFFT, and CTU

evaneschneider:
Deprecated CUFFT and PFFT gravity solvers; Deprecated CTU integrator; Added 1D and 2D simple integrators; Fixed bug in which characteristic tracing was not called for simple integrator in PPMP reconstruction.


bvillasen (approved):
Looks good!

PR #170, opened 2022-07-14
Add a std::vector like class for device arrays

bcaddy:
# Summary

- Passes all tests
- With the recent restructuring of branches I'm not sure where this PR should land. I've chosen `main` for now since I think we talked about CAAR being deprecated but please redirect it to the appropriate branch.

The new `cuda_utilities::DeviceVector` class provides a host side interface that is very similar to std::vector and includes Scope-Bound Resource Management/RAII for automatic allocation and freeing of device arrays. The class is intended to be used host side only and cannot be passed to kernels, however the `data()` method allows passing of the raw pointer to kernels[^1].

The new class is templatized and should support any primitive data type, it might support derived types as well but I haven't tested that.

## DeviceVector Methods

Detailed doxygen documentation can be found in `src/utils/DeviceVector.h`.

- `.data()` - returns the raw device pointer
- `.size()` - returns the number of elements in the array
- operator `[]` - overloaded indexing operator. Returns the value in the array at provided index. No bounds checking is performed
- `.at(size_t const &index)` - Returns the value in the array at provided index. Bounds checking is performed
- '.assign(T const &hostValue, size_t const &index=0)` - Assigns a value to location `index` in device array. Generally this should probably only be used for initializing scalar objects such as in reductions.
- `.resize(size_t const &newSize)` - Resizes the array. Previous values in the array are kept if they are still in range.
- `.reset(size_t const &newSize)` - Resizes the array. The contents of the array are uninitialized
- `.cpyHostToDevice(args)` - Copy values from host to device. This function is overloaded to accept either `std::vector` or raw pointer arrays and their size
- `.cpyDeviceToHost(args)` - Copy values from device to host. This function is overloaded to accept either `std::vector` or raw pointer arrays and their size

This commit also includes tests for the `cuda_utilities::DeviceVector` class.

[^1]: Making this class usable device side is not easy since it includes a lot device memory management methods and to use it device side and keep it synced with the host side version in memory would require lots of annoying and slow Memcpy's. Since the primary goal of the class is to provide easier memory management and host side access to values I deemed device side support unnecessary at this time.


bcaddy:
Of course! Generally it can be used to replace any explicit use of a device side pointer that points at an array or scalar value. The advantage of this class over just a raw pointer is that it will automatically free the allocated memory, provide easier host side access to the values in the device array, and make copying arrays to and from the host simpler. Here's an example for how I plan to use it in my function that calculates the magnetic divergence and a comparison to the current version that doesn't use it; it should be similarly usable for the integrators as well.

```cpp
// Current Version
Real launchCalculateMagneticDivergence(Real const *dev_conserved,
                                       Real const dx,
                                       Real const dy,
                                       Real const dz,
                                       int const nx,
                                       int const ny,
                                       int const nz,
                                       int const n_cells)
{
    // First let's create some variables we'll need and make sure they're allocated.
    static bool memoryAllocated=false;
    static Real *dev_maxDivergence, host_maxDivergence;
    static uint numBlocks, threadsPerBlock;
    if (not memoryAllocated)
    {
        CudaSafeCall( cudaMalloc(&dev_maxDivergence, sizeof(Real)) );
        reduction_utilities::reductionLaunchParams(numBlocks, threadsPerBlock);

        memoryAllocated = true;
    }

    // Now lets get the local maximum divergence
    hipLaunchKernelGGL(mhdUtils::calculateMagneticDivergence,
                        numBlocks, threadsPerBlock, 0, 0,
                        dev_conserved, dev_maxDivergence,
                        dx, dy, dz,
                        nx, ny, nz,
                        n_cells);
    CudaCheckError();
    CudaSafeCall(cudaMemcpy(&host_maxDivergence, dev_maxDivergence, sizeof(Real), cudaMemcpyDeviceToHost));
    cudaDeviceSynchronize();

    return host_maxDivergence;


// New version with DeviceVector
Real launchCalculateMagneticDivergence(Real const *dev_conserved,
                                       Real const dx,
                                       Real const dy,
                                       Real const dz,
                                       int const nx,
                                       int const ny,
                                       int const nz,
                                       int const n_cells)
{
    // First let's create some variables we'll need and make sure they're allocated.
    static DeviceVector dev_maxDivergence(1);
    // Currently reduction_utilities::reductionLaunchParams returns void. Now 
    // that we have C++17 I'll refactor it to return 2 values instead of passing 
    // in variables by reference.
    static const uint [numBlocks, threadsPerBlock] = reduction_utilities::reductionLaunchParams(numBlocks, threadsPerBlock);


    // Now lets get the local maximum divergence
    hipLaunchKernelGGL(mhdUtils::calculateMagneticDivergence,
                       numBlocks, threadsPerBlock, 0, 0,
                       dev_conserved, dev_maxDivergence.data(),
                       dx, dy, dz,
                       nx, ny, nz,
                       n_cells);
    CudaCheckError();
    cudaDeviceSynchronize();

    return dev_maxDivergence[0];
}
```

bcaddy:
HIP builds are fixed. I forgot to add new macros to convert from CUDA to HIP with the handful of new calls I added. Also, `cudaPointerAttributes` and `hipPointerAttribute_t` purport to be equivalent but `hipPointerAttribute_t` has slightly different member names and different values for the same meaning of those members, thus the ifdefs in the testing file

bcaddy:
I've rebased the `CAAR-deviceVector` off of the `dev` branch.

bcaddy:
Once PR #175 is merged I will rebase this PR so that the builds pass

PR #184, opened 2022-09-02
FFT and Cosmological ICs Generator

bvillasen:
This PR adds a class FFT3D and an initial condition generator for cosmological simulations.

New FFT3D objects can be created where needed and FFT filters can be easily implemented. See fft/fft_3D_filters.cu for examples. And see cosmology/cosmological_ics.cpp for an example of how to call the FFT filters.

To generate cosmological ICs the new parameters needed in the parameter file are:
init=Generate_Cosmological_ICs   
Omega_b=0.0497     //Omega Baryon
Init_redshift=100.0    //Initial redshift of the simulation
Init_temperature=282.023      //Initial temperature of the gas
cosmo_ics_random_seed=1234567   //Random seed (OPTIONAL, if omitted the system time will be used as seed)
cosmo_ics_pk_file=input_power_spectrum.txt  //file with the power spectrum to generate the dark matter and gas perturbations

A reference file for the input power spectrum for Plank_2018 Lamda-CDM + Eisenstein transfer function: extras/input_power_spectrum_lcdm.txt
An example parameter file to run a cosmological simulation with the ICs generator: parameters_cosmological_ics.txt

No compile flags are required to generate cosmological ICs.

Also this PR includes some cleaning of the chemistry_gpu module.


bcaddy:
I believe the reason that the builds are failing is that this PR is based off of an old version it the build script which is no longer working. I recommend rebasing off of `dev` and then targeting this PR to land on `dev` instead of `CAAR`

evaneschneider:
I think this PR has to stay in CAAR because the point is to make sure we can run these simulations for the FOM etc. - this is the last of the CAAR PRs. If it is possible to pull in the build changes from main that should fix the issue.

bcaddy:
Cherry picking commit `99c1659fb5a017cb95c903e88dea76eeed6e32c4` should do it; or just copy-paste the workflow file

PR #185, opened 2022-09-06
Fix crashes when there are zero local particles

alwinm:
When n_local = 0 and ngrid = 0, CUDA crashes as a result of trying to call a kernel with 0 blocks. Usually in the code this is avoided by always adding +1 to the number of blocks, but in my opinion I think it's preferable to avoid adding an extra block, avoid the kernel call altogether, and make the logic a little more explicit for future readers of the code. 

Currently I am proposing the pushed changes, but in the interest of consistency I can re-do these changes in the old way, or re-do the old ways to match these changes. 

I'll ping @ojwg since we've talked about these crashes, and I'll ping @bvillasen to tap his particles expertise.  

One nice feature about this pull request is that pure-hydro examples will now run with zero particles, even if particle flags were included.

### DIFF
--- a/src/particles/density_CIC_gpu.cu
+++ b/src/particles/density_CIC_gpu.cu
@@ -144,7 +144,7 @@ void Particles_3D::Get_Density_CIC_GPU_function(part_int_t n_local, Real particl
   dim3 dim1dBlock(TPB_PARTICLES, 1, 1);
 
   // Only runs if there are local particles
-  if (n_local) {
+  if (n_local > 0) {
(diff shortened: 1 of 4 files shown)


bcaddy:
I would replace `if (n_local)` with `if (n_local > 0)`. They should optimize to the same thing and the latter is clearer in my opinion.

alwinm:
After talking it out with Evan I think the rest of the ngrid logic is fine as is and this doesn't introduce a major inconsistency so in my opinion this pull request is good to go.

PR #192, opened 2022-09-22
Allow for chemistry units to work with non-cosmological simulations.

### LINKED ISSUE #190: Chemistry functions require COSMOLOGY, needs minor re-writing.
Nick reports that:

There are also non-cosmological values that are not defined:
src/chemistry_gpu/chemistry_functions.cpp: In member function ‘void Grid3D::Compute_Gas_Temperature(Real*, bool)’:
src/chemistry_gpu/chemistry_functions.cpp:257:31: error: ‘n_cells’ was not declared in this scope
  257 |         GE = (dev_conserved[4*n_cells + id] - 0.5*d*(vx*vx + vy*vy + vz*vz));
Does the CHEMISTRY_GPU actually work at all?

I also encountered compile errors while trying compile with CHEMISTRY_GPU set:
src/chemistry_gpu/chemistry_functions.cpp: In member function ‘void Grid3D::Initialize_Chemistry(parameters*)’:
src/chemistry_gpu/chemistry_functions.cpp:35:18: error: ‘struct parameters’ has no member named ‘H0’
   35 |   Chem.H.H0 = P->H0;
      |                  ^~

brantr:
Minor adjustments were made to chemistry_functions.cpp and chemistry_functions_gpu.cu to enable non-cosmological calculations.

The cosmological unit conversions have been moved into COSMOLOGY ifdef logic. I believe the unit system is CGS, with density (rho_0), factors of a, and factors of h applied after the CGS units are initially defined.

There was also an adjustment replacing a call to dev_conserved with a call to total energy E that avoids referencing n_cells in chemistry_functions.cpp.

This needs to be tested with an RT test compilation and calculation. So I'm assigning to @evaneschneider but I suspect Nick needs to test since he raised the issue.

Would close #190.

### DIFF
--- a/src/chemistry_gpu/chemistry_functions.cpp
+++ b/src/chemistry_gpu/chemistry_functions.cpp
@@ -31,30 +31,45 @@ void Grid3D::Initialize_Chemistry( struct parameters *P ){
   Chem.H.Temp_end   = 1000000000.0;
   
   Chem.H.H_fraction = INITIAL_FRACTION_HI + INITIAL_FRACTION_HII;
-  
+ 
+#ifdef COSMOLOGY 
   Chem.H.H0 = P->H0;
   Chem.H.Omega_M = P->Omega_M;
   Chem.H.Omega_L = P->Omega_L;
-  
+#endif //COSMOLOGY
   
   // Set up the units system.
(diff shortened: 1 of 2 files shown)


evaneschneider (approved):
This looks good to me, but Nick will have to test.

Issue #189, opened 2022-09-14
No error message for missing config file


bcaddy:
If you try to run Cholla with a non-existent config file you get complex PMIx errors instead of just a error message. There should be some checking that a config file exists and that it has the proper info. C++17 makes this much easier with `std::filesystem`.

Example:

```
rcaddy@c3po:bin$ mpirun -n 4 ./cholla.hydro.c3po ./sod512.txt
No protocol specified
Git Commit Hash = 646614bd94fd580c2b09a2a11a1e7a9c488427bd
Macro Flags     = -DCUDA -DMPI_CHOLLA -DPRECISION=2 -DPPMC -DHLLC -DSIMPLE -DDENSITY_FLOOR -DTEMPERATURE_FLOOR -DOUTPUT -DHDF5 -DGIT_HASH=646614bd94fd580c2b09a2a11a1e7a9c488427bd
Parameter values:  nx = 0, ny = 0, nz = 0, tout = 0.000000, init = , boundaries = 0 0 0 0 0 0
Output directory:  pRs��

Creating Log File: run_output.log

nproc_x 2 nproc_y 2 nproc_z 1
Allocating MPI communication buffers on GPU (nx = 0, ny = 0, nz = 1280).
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 3 in communicator MPI_COMM_WORLD
with errorcode -1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
[c3po:912606] PMIX ERROR: UNREACHABLE in file ../../../src/server/pmix_server.c at line 2193
[c3po:912606] PMIX ERROR: UNREACHABLE in file ../../../src/server/pmix_server.c at line 2193
[c3po:912606] PMIX ERROR: UNREACHABLE in file ../../../src/server/pmix_server.c at line 2193
Local number of grid cells: 0 0 0 512
Setting initial conditions...
ABORT: : Unknown initial conditions!
[c3po:912606] 3 more processes have sent help message help-mpi-api.txt / mpi-abort
[c3po:912606] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

PR #193, opened 2022-09-23
Updating Cloudy Cooling for CUDA/hip texture objects

alwinm:
- Cholla checks two common paths (current working directory, and src/cooling/) for the cloudy table, and upon success chprintf's the relative path of the table it loads, or upon failure chprintf's the relative paths it expected the table to be, for easier future debugging. 
- Cloudy textures updated to be texture objects, which need to be passed into device functions. Texture interfaces updated to avoid unexpected numerical errors due to texture coordinates and interpolation, with comments added where appropriate. 
- gpu.hpp adds Cuda to hip translations for texture-related definitions
- cooling_wrapper.* renamed to load_cloudy_texture.*. Can consider renaming Load_Cuda_Texture() and Free_Cuda_Texture() to Load_Cloudy_Texture() and Free_Cloudy_Texture() in the future, in particular if there are non-Cloudy textures in the future. 
- texture_utilities.h added to easily share basic texture interpolation device functions
- make.type.cloudy added for ease of use


alwinm:
Made a light improvement with regards to this: https://github.com/cholla-hydro/cholla/issues/189
I figure if the param file fails to load, it should be fine to exit cholla at that step. 

Tested with regression tests of type hydro and cooling, which made me realize I needed to give the textures an initial value in cooling_cuda.cu rather than load_cloudy_textures. The alternative would be an placing ifdef around the kernel call to cooling_kernel, so that when CLOUDY is off, 0s are passed instead of the texture object.

PR #200, opened 2022-10-10
new build-tests.yml for Docker container

alwinm:
I set up this pull request by cherry-picking the dev commit, so down the road if/when dev gets merged into main, this won't generate a merge conflict!

I set up the build script to look for the docker containers: 

docker://alwinm/cholla:hip_github
docker://alwinm/cholla:cuda_github

When we create new versions, we can give them tags with other names, and then alias the tagname to hip_github and cuda_github. This way, we can update the Docker container while keeping a version history and easily being able to switch back to previous versions if necessary, all without requiring pull requests on github. 

I can set maintainers as collaborators on the Docker Hub repo so any of us can push images! 

Example:

Setting the github container version to 0.0, assuming you've already pushed 0.0.
```
docker tag alwinm/cholla:hip_v0.0 alwinm/cholla:hip_github
docker push alwinm/cholla:hip_github
```

Setting the github container version to 0.1, assuming you've already pushed 0.1.
```
docker tag alwinm/cholla:hip_v0.1 alwinm/cholla:hip_github
docker push alwinm/cholla:hip_github
```

### DIFF
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -58,7 +58,7 @@ jobs:
     steps:
 
     # Install required Tools
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
     # Show versions
     - name: Show MPI version
--- a/.github/workflows/build_tests.yml
+++ b/.github/workflows/build_tests.yml
@@ -2,7 +2,6 @@ name: Cholla Compile
 
 on:
   pull_request:
-#    branches: [ CAAR ]
   schedule:
     - cron: "37 07 * * 1"  # run every Monday at 07:37UTC. Crontab computed with crontab.guru
   workflow_dispatch:
@@ -11,31 +10,37 @@ jobs:
   Build:
     name: >
       Build
-      API:${{ matrix.gpu-api }}
-      Make-Type:${{ matrix.make-type }}
-      Cuda-toolkit:v${{ matrix.cuda-toolkit-version }}
-      GCC:v${{ matrix.gcc-version }}
-      ROCm:v${{ matrix.rocm-version }}
+      ${{ matrix.container.name }}
+      TYPE=${{ matrix.make-type }}
+
+    # Cuda-toolkit:v${{ matrix.cuda-toolkit-version }}
+    # GCC:v${{ matrix.gcc-version }}
+    # ROCm:v${{ matrix.rocm-version }}
     # if: ${{ false }}  # If uncommented this line will disable this job
 
     # Choose OS/Runner
     runs-on: ubuntu-latest
-
+    container: 
+      image: ${{matrix.container.link}} 
+    defaults:
+      run:
+        shell: bash
     # Matrix for different make types
     strategy:
       fail-fast: false
       matrix:
         make-type: [hydro, gravity, disk, particles, cosmology, mhd]
-        gpu-api: [HIP, CUDA]
+        container: [{name: "CUDA", link: "docker://alwinm/cholla:cuda_github"}, {name: "HIP",link: "docker://alwinm/cholla:hip_github"},]
+        # gpu-api: [CUDA]
         # NOTE: if more than one parameter is in any of these three variables
         # you need to manually exclude it for the GPU API that doesn't use it.
         # An example exclude is shown below but commented out. Uncomment and
         # tweak it to fit your needs
         # CUDA uses the cuda-toolkit-version and gcc-version
         # HIP uses the clang-version
-        cuda-toolkit-version: ['11.2.2']
-        gcc-version: [9]
-        rocm-version: ['5.1.0']
+        # cuda-toolkit-version: ['11.2.2']
+        # gcc-version: [9]
+        # rocm-version: ['5.1.0']
         mpi: ['openmpi'] #Can use mpich and/or openmpi
         # exclude:
         #   - gpu-api: HIP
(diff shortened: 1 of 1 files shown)


.github/workflows/build_tests.yml
```diff
@@ -1,165 +1,110 @@
...
+        # NOTE: if more than one parameter is in any of these three variables
+        # you need to manually exclude it for the GPU API that doesn't use it.
+        # An example exclude is shown below but commented out. Uncomment and
+        # tweak it to fit your needs
+        # CUDA uses the cuda-toolkit-version and gcc-version
+        # HIP uses the clang-version
+        # cuda-toolkit-version: ['11.2.2']
+        # gcc-version: [9]
+        # rocm-version: ['5.1.0']
+        mpi: ['openmpi'] #Can use mpich and/or openmpi
```
bcaddy, line 44: I don't think this like is required anymore and the other old matrix stuff should be removed rather than commented out

.github/workflows/build_tests.yml
```diff
@@ -1,165 +1,110 @@
...
+    # Install required Tools
+    - uses: actions/checkout@v3
+
+    # Show versions
+    - name: Show MPI version
+      run: mpirun --version
+    - name: Show HDF5 config
+      run: |
+        h5cc -showconfig
+    - name: Git Safe Directory
```
bcaddy, line 69: I'm curious, what does this do?

bcaddy (changes requested):
This looks great. I left a few comments but nothing major.

The reason I marked this as requesting changes is:

1. We absolutely need how to rebuild the docker container documented somewhere. I would prefer a script or GHA workflow that would do it.
2. I actually think that requiring a PR to update the version of the container we're using is a good idea. It makes it clear exactly which container was used, when we switched between them.

Side note. Would you be willing to give the group a talk/workshop on containers? Maybe 10/21? We can record if for those not able to attend.

bcaddy:
Triggering a workflow to rebuild the containers would be easy to do with labels as described [here](https://blog.mergify.com/running-github-actions-only-on-certain-pull-requests/)

evaneschneider:
@bcaddy - does the documentation on the wiki seem sufficient, or are you suggesting that we need more?

.github/workflows/build_tests.yml
bcaddy: Thanks for the info.

bcaddy:
I didn't see that, sorry. It looks good to me.

I still think that changing the container that the builds use should require a PR

alwinm:
Currently the documentation I have so far is here: https://github.com/cholla-hydro/cholla/wiki/Docker

It takes 1 command to build a Docker image, 1 command to run it (I always do this to make sure Cholla compiles on it before pushing) 1 command to tag it (naming it whatever you want), and 1 command to push it. We would need to automate the step that verifies that Cholla will compile.

I could definitely do a short tutorial for the group on Docker.

I have some example Dockerfiles here:
https://github.com/alwinm/cholladocker

evaneschneider:
Doesn't changing the docker container require a PR by default? Since it involves changing the version number in a file?

bcaddy:
If my current understanding is correct then if doesn't since the Cholla workflow doesn't have a version number. Rather someone builds a new container and then aliases it to the name that the workflow file is looking for.

alwinm:
Working under the assumption that you preferred fewer PRs required, I actually stripped out version numbers in build-tests.yml. Someone who changes the Docker container would have to change version numbers in that Dockerfile, but currently there are no Dockerfiles being tracked in the cholla git repo. And any collaborators of the alwinm/cholla Docker hub repo can push / change the container.

bcaddy:
My suggestIon is that the docker file be tracked and a GHA is written to rebuild the container when needed and that we don't test the container in that GHA since the builds are run on every PR so the testing will happen immediately after always.

So the workflow looks like:

1. Contributor forks and branches
2. Modifies dockerfile and build-tests.yml
3. Makes a PR with an appopriate label to tell github to launch the docker rebuild
4. Once it's done we either manually or automatically run the cholla build to test the containers

It might be easier to have a section of build-tests.yml that checks if the container exists and if not it starts the build and waits for it to complete

bcaddy:
Also, it might be better to store the containers in a "cholla" account rather than @alwinm's personal one long term. That's easy to change later though

alwinm:
I can see some of the virtues of that approach, but I'm not sure they outweigh the labor cost. 

100% agree we should have a cholla account one day.

bcaddy:
> I can see some of the virtues of that approach, but I'm not sure they outweigh the labor cost.

I agree, that's why I think we should automate it. I'm still looking but I think we can just have build-tests.yml check if the container exists and if not the build it. This should be done in a pre-req job so every single build type doesn't try to make a new container.

These links should do it:
- [check if container exists](https://stackoverflow.com/questions/60892354/how-to-check-if-a-version-of-a-github-docker-image-already-exists)
- [build container in GHA](https://event-driven.io/en/how_to_buid_and_push_docker_image_with_github_actions/)
- [GHA pre-req jobs](https://lannonbr.com/blog/github-actions-jobs/)

evaneschneider:
Given that this version compiles, passes all the build checks, and is significantly faster, and the impetus was to get it in ASAP so that @ojwg can move ahead with PR 199, I'd like to merge it as-is, and we can work out the details of how to implement the docker container with version reference later. @alwinm, do these docker containers have the necessary libraries for Orlando's new code?

evaneschneider (approved):

PR #213, opened 2022-12-02
Grid Enum - dev

alwinm:
Provides an enum object to help with accessing different combinations of scalar fields and adding new scalar/non-scalar fields in the future. enum object automatically calculates NSCALARS based on what is enrolled, and grid_enum::nfields should be the number of fields in grid.

Short-term goal is to allow scalars to be numbered more flexibly. 

```#ifdef SCALAR ``` will now activate scalars in general. in order to use what used to be scalar0, use 
```#ifdef BASIC_SCALAR```

Currently many areas of the code are consistent with this approach but NOT explicitly moved over. e.g. accessing ```momentum_x``` is **currently written** as ```C.device[1*n_cells]``` but **would be the same** as ```C.device[n_cells*grid_enum::momentum_x]```. Same holds for magnetic fields, gas energy, etc. 

Currently nfields throughout the code is NOT using grid_enum::nfields.

@helenarichie please bring grid_enum into DUST and complain to me if you are unable to commit directly to the branch or make a pull request into https://github.com/alwinm/cholla/tree/enum-dev. We should not merge this PR until we make sure DUST is consistent, the other scalars have been converted already.

### CODE REVIEW

**Comment by bcaddy** [comment]
I like this but have 2 questions

1. Can this be used in device code? 
2. Could we move the enum to `global.h` instead of `grid` since it probably needs to be accessed globally

**Comment by bcaddy** [comment]
I think the `global.h` vs. `grid3d.h` debate might be more a symptom of how the `Grid3D` class doesn't just do grid stuff and the actual grid stuff is spread across `Grid3D`, `Conserved`, and `Header`. Having it where it currently is is fine though I think we should have a discussion about a major refactor of the grid vs. integrator/simulation code (or replace the grid with an ECP library??).

File: src/grid/initial_conditions.cpp [hunk 1]
```diff
@@ -1335,8 +1351,16 @@ void Grid3D::Clouds()
             #ifdef DE
             C.GasEnergy[id]  = p_cl/(gama-1.0);
             #endif
+
+            #ifdef SCALAR
+	    #ifdef BASIC_SCALAR
+            C.basic_scalar[id] = C.density[id]*0.3;
+	    #endif
+
+	    //TODO for Helena: Alwin merged this from dev but isn't sure if it is right.
```
alwinm, line 1360: @helenarichie I wasn't sure what I was doing when I merged in this line. Can you check to make sure that it's right or if it needs to be changed to access the dust field in particular?

helenarichie: This has been resolved in my latest commit to Alwin's enum-dev branch! All tests for hydro and dust pass.

src/grid/grid3D.cpp
```diff
@@ -327,12 +334,12 @@ void Grid3D::AllocateMemory(void)
 
 
   #ifdef CHEMISTRY_GPU
-  C.HI_density    = &C.scalar[ 0*H.n_cells ];
-  C.HII_density   = &C.scalar[ 1*H.n_cells ];
-  C.HeI_density   = &C.scalar[ 2*H.n_cells ];
-  C.HeII_density  = &C.scalar[ 3*H.n_cells ];
-  C.HeIII_density = &C.scalar[ 4*H.n_cells ];
-  C.e_density     = &C.scalar[ 5*H.n_cells ];
+  C.HI_density     = &C.density[ H.n_cells*grid_enum::HI_density ];
```
evaneschneider, line 337: Would it be possible / make more sense to lines like this to reference C.host instead of C.density? I think one of the goals with this approach is to move away from having things point to arrays with confusing names, and since the creation of C.host and C.device, I don't actually know that we need the old conserved array pointers any more.

alwinm: Great point--agreed! How many of those changes do you want included in this PR specifically? Since changing all calls everywhere would be substantially more work.

FYI I started a discussions thread where I am listing potential to-dos (for a hacking day or otherwise) related to enum. 

https://github.com/cholla-hydro/cholla/discussions/215

bcaddy:
Question/idea.

With this enum method we would index fields like `[idx + enumValue * n_cells]`. Is there a way to just multiply all the enum values by `n_cells` so we don't have to include that too?

Also, the magnetic fluxes, interface states, and CT electric fields don't follow the same rules as everything else to save memory. Can there be items in the enum with duplicate values or should I just have my own method for dealing with those oddities? Maybe a magnetic enum?

src/grid/grid3D.cpp
evaneschneider: I think for this PR just using C.host or C.device where you've already changed them is fine. We can work on making it more consistent through the code base later.

evaneschneider:
Actually I had a question for you about that Bob. Are the values that are currently contained in the three magnetic field arrays the cell-averaged values, or the face-centered values? Because if it's the latter then I think that multiplying by n_cells is not correct, since there is an extra face, right?

bcaddy:
They're face centered. Multiplying by n_cells is still correct. Since theres an extra grid cell already for other stuff the actual size of the arrays are all the same. The staggeredness of the grid comes in when there's occasional `+0.5*dx` floating around

evaneschneider:
Right now I think none of the temporary fields are in the grid_enum header, right? So maybe that's something we can discuss tomorrow and deal with in a future PR?

bcaddy:
As I understand it the grid_enum just provides a bunch of integers for us to index off of. So they should work fine for all the temporary fields since they have the same ordering.

Followup question. Why is the grid one giant array? and not a bunch of individual arrays for each field or a class with a bunch of pointers? i.e. why isn't the C class our primary way of interacting with the grid?

evaneschneider:
I think this will be easier to talk out / explain in person. Let's put it on the agenda for tomorrow.

alwinm:
> Is there a way to just multiply all the enum values by `n_cells` so we don't have to include that too?

As I understand it, n_cells is set at run-time and enums are set at compile time and const which is helpful in multiple ways, so I don't know how to accomplish that. Unclear to me whether accessing a run-time computed variable would hit performance or be a pain to pass to the kernel. 

> Also, the magnetic fluxes, interface states, and CT electric fields don't follow the same rules as everything else to save memory. Can there be items in the enum with duplicate values or should I just have my own method for dealing with those oddities? Maybe a magnetic enum?

If those are not held in C.host/C.device it could make sense for them to live in a magnetic enum. There can be items in the enum with duplicate values, which you could set like this:

```
a,
b,
b2 = b,
c,
c2 = c,
d,
```

Or if it's more readable for you
```
a,
b,
c,
d,
// Aliases
b2 = b,
c2 = c,
```

My understanding of the giant array is to ensure memory is allocated in a smart way.

Issue #219, opened 2022-12-20
Dust make type doesn't build on Crusher


bcaddy:
As part of testing the MHD code I've been running all the tests on AMD and NVIDIA. The dust build type won't build at all on AMD systems and throws this error

```
hipcc -std=c++17 --offload-arch=gfx90a -Wno-unused-result -DCUDA -DMPI_CHOLLA -DBLOCK -DPRECISION=2 -DPPMC -DHLLC -DAVERAGE_SLOW_CELLS -DTEMPERATURE_FLOOR -DVL -DSCALAR -DDUST -DCOOLING_GPU -DSLICES -DPROJECTION -DOUTPUT -DHDF5 -DMPI_GPU -DO_HIP -DGIT_HASH='"01e6cf2630ae9e0951e47c3e6e421a8d7428326b"' -DMACRO_FLAGS='"-DCUDA -DMPI_CHOLLA -DBLOCK -DPRECISION=2 -DPPMC -DHLLC -DAVERAGE_SLOW_CELLS -DTEMPERATURE_FLOOR -DVL -DSCALAR -DDUST -DCOOLING_GPU -DSLICES -DPROJECTION -DOUTPUT -DHDF5 -DMPI_GPU -DO_HIP -DGIT_HASH='"01e6cf2630ae9e0951e47c3e6e421a8d7428326b"'"' -Isrc -I/opt/cray/pe/hdf5/1.12.1.5/crayclang/14.0/include -I/opt/cray/pe/mpich/8.1.17/ofi/crayclang/10.0/include -Wall -ferror-limit=1 -c src/dust/dust_cuda.cu -o src/dust/dust_cuda.o
src/dust/dust_cuda.cu:114:17: error: __device__ function 'calc_tau_sp' cannot overload __host__ __device__ function 'calc_tau_sp'
__device__ Real calc_tau_sp(Real n, Real T) {
                ^
src/dust/dust_cuda.h:15:26: note: previous declaration is here
__device__ __host__ Real calc_tau_sp(Real n, Real T);
                         ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
2 errors generated when compiling for gfx90a.
make: *** [Makefile:189: src/dust/dust_cuda.o] Error 1
```

alwinm:
https://github.com/bcaddy/cholla/pull/14

Since I've seen this kind of thing before I made a PR to fix it--just needs to add __host__ consistently

evaneschneider:
[...]

Issue #218, opened 2022-12-20
Spherical Collapse gravity test currently failing


bcaddy:
The `tGRAVITYSYSTEMSphericalCollapse.CorrectInputExpectCorrectOutput` and `tPARTICLESSYSTEMSphericalCollapse.CorrectInputExpectCorrectOutput` system tests are currently failing on `dev`. Tested on C-3PO

evaneschneider:
PR #223 should fix this.

bcaddy:
It looks like PR #223 fixed the gravity test but not the particles test

PR #220, opened 2022-12-20
Add Preliminary MHD Integrator

bcaddy:
All non-MHD tests pass except those noted in issues #218 and #219 which also don't pass in `dev`.

This commit primarily adds support for MHD to the Van Leer 3D integrator along with myriad small fixes and modifications. All new features have tests as well.

The MHD implementation still has bugs but needs to be merged in before we reformat the code so that the merging doesn't provoke too many merge conflicts. 

# MHD

- Add MHD support to Van Leer 3D integrator
- All MHD code is in the `mhd` namespace with sub-namespaces when relevant (e.g. `mhd::utils`)
- Add kernel to compute the CT electric fields and tests for it
- Add kernel to update the magnetic field and tests for it
- Add MHD support to the PCM reconstruction
- Add new `Linear_Wave` initial condition type
- Add `ctElectricFields` global device array
- Set all MHD input parameters to default to zero
- Remove incorrect MHD specializations in the boundary conditions
- Add check to make sure MHD has at least 3 ghost cells
- Add functions to compute the maximum magnetic divergence, report it, and exit if it's too large
- Clean up make.type.mhd and enable cuda error checking
- Add example parameter files for
  - Alfven wave
  - Fast magnetosonic wave
  - Slow magnetosonic wave
  - Ryu & Jones 1995 shock tubes 1a and 2a
- Add the following input parameters for generalized linear waves
  - EigenVec_rho
  - EigenVec_MomentumX
  - EigenVec_MomentumY
  - EigenVec_MomentumZ
  - EigenVec_E
  - EigenVec_Bx
  - EigenVec_By
  - EigenVec_Bz
  - pitch
  - yaw
- Added MHD system tests for:
  - Hydro only constant state test
  - MHD constant state
  - Sod test that works with PCM (this will be merged with the hydro test when MHD supports PPMC)
  - All 4 MHD linear waves
  - MHD Einfeldt strong rarefaction

# HLLD Riemann Solver

- Updated interface state indexing for new format
- Added documentation
- Moved dot product function to `math_utils`

# Testing

- Add new method, `runL1ErrorTest`, to `SystemTestRunner` that computes the L1 error compared to the initial conditions. Ideal for wave tests
- Updated hydro wave tests to use new `runL1ErrorTest` method

# Reductions

- Add new FP atomic max that works on CUDA or ROCm
- Update time step calculation to use the grid reduction with atomic and DeviceVector
- Updated tests to match
- Remove dev_dti, host_dti_array, and dev_dti_array global variables

# Utilities

## Cuda Utilities

- Add function `initGpuMemory` which initializes GPU memory so the CUDA compute sanitizer doesn't complain about it. Used to initialize all GPU arrays that I know of.
- Add struct `AutomaticLaunchParams` which is a thin wrapper over the occupancy API. Primarily intended for reductions where performance is sensitive to the number of blocks but could be used for any kernel launch

## DeviceVector

- New option to initialize memory

## Math Utilities

New namespace and file for math utilities. Currently contains a semi-general rotation and dot product functions

## MHD Utilities

- The `mhd::utils::computeEnergy` function now works properly with either MHD or hydro and returns the appropriate energy

# Other

- Add a function `to_string_exact` to convert floating point numbers to a string such that it can be exactly deserialized back from a string to the same floating point number. Only used in tests currently
- Commit [379b86e](https://github.com/cholla-hydro/cholla/pull/220/commits/379b86e15def7090e5aeceb95bb5c2ea81bd8148) in this PR should Fix Issue #219 as well.


bcaddy:
This is currently a draft and not ready for merging but wanted to make the changes available for code review as the things that are left to do are minor and should be done shortly.

# To do

- [ ] Rebase off of `dev` again now that PR #213 is merged
- [ ] Replace current indexing in MHD code with grid enums

bcaddy:
Documentation for the `runL1Error` method has been added to the wiki

Makefile
```diff
@@ -38,6 +38,9 @@ ifeq ($(TEST), true)
   CFLAGS   += $(TEST_FLAGS)
   CXXFLAGS += $(TEST_FLAGS)
   GPUFLAGS += $(TEST_FLAGS)
+  CFLAGS   += $(TEST_FLAGS)
```
evaneschneider, line 41: Accidental duplicates here?

Makefile
```diff
@@ -134,6 +137,7 @@ ifdef HIPCONFIG
   DFLAGS    += -DO_HIP
   CXXFLAGS  += $(HIPCONFIG)
   GPUCXX    ?= hipcc
+  GPUFLAGS  += -Wall -ferror-limit=1
```
evaneschneider, line 140: What does this do?

src/global/global_cuda.cu
```diff
@@ -9,16 +9,12 @@
 bool memory_allocated;
 Real *dev_conserved, *dev_conserved_half;
 Real *Q_Lx, *Q_Rx, *Q_Ly, *Q_Ry, *Q_Lz, *Q_Rz, *F_x, *F_y, *F_z;
+Real *ctElectricFields;
 Real *eta_x, *eta_y, *eta_z, *etah_x, *etah_y, *etah_z;
-Real *dev_dti;
```
evaneschneider, line 13: Is the time step reduction fixed so that we don't need this any more?

bcaddy:
Converting MHD to using `grid_enum` is done so I'm converting this from a draft to a full PR. 

@alwinm, switching to using the enum was pretty easy. Could you take a look at my additions to the enum and my general usage? I think it's what we talked about but I'd like confirmation and feedback.

Makefile
bcaddy: Yep. Fixing. 

There were a few spots like that that got duplicated in the rebase. I thought I got them but if you see any others please let me know.

Makefile
bcaddy: Turn on all warnings and stop compiling after one error. I wanted more warnings for debugging. I'll turn off the error limit

src/global/global_cuda.cu
bcaddy: The time step reduction is fixed. I replace the atomic with the built in one on HIP and some fancy bit casting shenanigans on CUDA. The CUDA atomic is taken from CUML.

src/grid/grid_enum.h
```diff
@@ -69,7 +69,29 @@ enum : int {
   num_fields,
 
 //Aliases and manually computed enums
+  #ifdef  MHD
```
alwinm, line 72: This looks good! 

Adding more comments 

- Wrap the magnetic fluxes in ifdef MHD so that it compiles for non-MHD builds 
- You have the option of doing something like
```
  fluxX_magnetic_z = magnetic_start,
  fluxX_magnetic_y, // instead of  fluxX_magnetic_y = magnetic_start+1,
```
but this is a case where I imagine you're preferring explicit over implicit
- I have a slight preference for moving ```nscalars = finalscalar_plus_1 - scalar,``` to the top of the list since everything else is being wrapped by ifdef MHD but NSCALARS always needs to be defined. 



```
// Aliases and manually computed enums
  nscalars = finalscalar_plus_1 - scalar, // Prefer this to be at the top of list or at the end
  
  #ifdef  MHD
  num_flux_fields = num_fields-1,
  num_interface_fields = num_fields-1,
  #else
  num_flux_fields = num_fields,
  num_interface_fields = num_fields,
  #endif  //MHD

  #ifdef MHD
  magnetic_start = magnetic_x,
  magnetic_end   = magnetic_z, // Unused, consider 
  // Note that the direction of the flux, the suffix _? indicates the direction of the electric field, not the magnetic flux
  fluxX_magnetic_z = magnetic_start,
  fluxX_magnetic_y = magnetic_start+1,
  fluxY_magnetic_x = magnetic_start,
  fluxY_magnetic_z = magnetic_start+1,
  fluxZ_magnetic_y = magnetic_start,
  fluxZ_magnetic_x = magnetic_start+1,
  Q_x_magnetic_y = magnetic_start,
  Q_x_magnetic_z = magnetic_start+1,
  Q_y_magnetic_z = magnetic_start,
  Q_y_magnetic_x = magnetic_start+1,
  Q_z_magnetic_x = magnetic_start,
  Q_z_magnetic_y = magnetic_start+1
  #endif
```

bcaddy: - I'm in the middle of fixing it for non-mhd builds now
- I'd prefer explicit. Also, IMO everything in the "aliases" section should be explicit since that's the section where we're saying ordering doesn't matter
- I moved `nscalars = finalscalar_plus_1 - scalar,`

The commit for all of this should be up shortly

alwinm: If you like you can also move these over. This is just the output of ```git grep -ip +[\s]*NSCALARS```

Just wanted to point it out but no worries if it's not a priority. 

```
src/mhd/ct_electric_fields_tests.cu=public:
src/mhd/ct_electric_fields_tests.cu:        grid (n_cells * (8+NSCALARS)),
src/mhd/magnetic_update_tests.cu=public:
src/mhd/magnetic_update_tests.cu:        sourceGrid      (n_cells * (8+NSCALARS)),
src/mhd/magnetic_update_tests.cu:        destinationGrid (n_cells * (8+NSCALARS), -999.),
src/mhd/magnetic_update_tests.cu:        fiducialData    (n_cells * (8+NSCALARS), -999.),
src/utils/mhd_utilities_tests.cu=TEST(tMHDCellCenteredMagneticFields,
src/utils/mhd_utilities_tests.cu:    std::vector<double> testGrid(n_cells * (8+NSCALARS));
```

bcaddy: Those are all tests and aren't actually indexing, they're creating a mock grid for testing. I've replaced them with `grid_enum::num_fields` and it appears to be working. There's a few other uses floating around in the tests, and only there as far as I can tell, and I replaced them with `grid_enum::nscalars` since they're actually counting scalars and not indexing for the magnetic field

bcaddy: I also didn't touch any of the scalar loops in PCM or the HLLD solver. I don't know enough about scalars to risk breaking something.

alwinm: Raw NSCALARS (no plus signs) can be left as is or replaced, they are being macro-replaced with grid_enum::nscalars anyway. grid_enum is designed to maintain compatibility with the scalar loops throughout the code by correctly calculating NSCALARS.

bcaddy: Sounds good. I'm going to mark this conversation as resolved, feel free to reopen it if needed.

bcaddy:
FYI, I think github actions are having some issues at the moment. Some jobs are getting shutdown signals from somewhere. I double checked and everything builds fine on C-3PO and the containers running on my machine. For now I'm just restarting failed jobs until they pass.

src/hydro/hydro_cuda.cu
```diff
@@ -1207,12 +1193,12 @@ __device__ void Average_Cell_All_Fields( int i, int j, int k, int nx, int ny, in
   Average_Cell_Single_Field( 4, i, j, k, nx, ny, nz, ncells, conserved );
   #ifdef  MHD
     // Average MHD
-    Average_Cell_Single_Field( 5+NSCALARS, i,   j,   k,   nx, ny, nz, ncells, conserved );
-    Average_Cell_Single_Field( 6+NSCALARS, i,   j,   k,   nx, ny, nz, ncells, conserved );
-    Average_Cell_Single_Field( 7+NSCALARS, i,   j,   k,   nx, ny, nz, ncells, conserved );
-    Average_Cell_Single_Field( 5+NSCALARS, i-1, j,   k,   nx, ny, nz, ncells, conserved );
-    Average_Cell_Single_Field( 6+NSCALARS, i,   j-1, k,   nx, ny, nz, ncells, conserved );
-    Average_Cell_Single_Field( 7+NSCALARS, i,   j,   k-1, nx, ny, nz, ncells, conserved );
+    Average_Cell_Single_Field( grid_enum::magnetic_x, i,   j,   k,   nx, ny, nz, ncells, conserved );
```
evaneschneider, line 1196: I'm actually curious about how this works. Does the magnetic field averaging violate the divergence rule? This may be more complicated than the way it is currently written.

src/integrators/VL_3D_cuda.cu
```diff
@@ -166,29 +225,46 @@ void VL_Algorithm_3D_CUDA(Real *d_conserved, Real *d_grav_potential, int nx, int
...
+    // Step 5.5: Compute the Constrained transport electric fields
+    hipLaunchKernelGGL(mhd::Calculate_CT_Electric_Fields, dim1dGrid, dim1dBlock, 0, 0, F_x, F_y, F_z, dev_conserved_half, ctElectricFields, nx, ny, nz, n_cells);
+    CudaCheckError();
+    #endif  //MHD
 
     // Step 6: Update the conserved variable array
     hipLaunchKernelGGL(Update_Conserved_Variables_3D, dim1dGrid, dim1dBlock, 0, 0, dev_conserved, Q_Lx, Q_Rx, Q_Ly, Q_Ry, Q_Lz, Q_Rz, F_x, F_y, F_z, nx, ny, nz, x_off, y_off, z_off, n_ghost, dx, dy, dz, xbound, ybound, zbound, dt, gama, n_fields, density_floor, dev_grav_potential);
     CudaCheckError();
 
+    #ifdef  MHD
```
evaneschneider, line 251: Does the MHD update care about the conserved variable update at all? If so, it should be after the dual energy update. If not, it doesn't matter. I'm just not 100% sure about the ordering here.

evaneschneider (commented):
I've only made it through the integrator, but figured I'd leave what comments I have so far.

src/grid/initial_conditions.cpp
```diff
@@ -291,6 +297,61 @@ void Grid3D::Sound_Wave(Real rho, Real vx, Real vy, Real vz, Real P, Real A)
...
+
+        // set constant initial states. Note that mhd::utils::computeEnergy
+        // computes the hydro energy if MHD is turned off
+        Real sine_wave = std::sin(2.0 * PI * x_pos);
+
+        C.density[id]    = rho;
+        C.momentum_x[id] = rho*vx;
+        C.momentum_y[id] = rho*vy;
+        C.momentum_z[id] = rho*vz;
+        C.Energy[id]     = mhd::utils::computeEnergy(P, rho, vx, vy, vz, Bx, By, Bz, gama);
```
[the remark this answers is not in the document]
bcaddy: Only this function is compiled if MHD is turned off. `mhd::utils::computeEnergy` returns the correct hydro or mhd energy depending on what flags are enabled. IMO we really need to unify these utility functions. We should be able to just call "computeEnergy" or something similar and it will automatically choose the correct way to compute the energy, then there can be separate functions if you need a specific one. Otherwise at some point someone is going to be computing the hydro energy when they want the MHD energy or vice versa

src/hydro/hydro_cuda.cu
bcaddy: No clue. I don't think it does but I haven't checked

src/integrators/VL_3D_cuda.cu
bcaddy: It doesn't. It only uses the CT electric fields and previous magnetic fields then writes out the updated fields.

src/grid/grid3D.cpp
```diff
@@ -603,6 +609,39 @@ void Grid3D::Update_Time(){
 
 }
 
+#ifdef  MHD
```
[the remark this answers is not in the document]
bcaddy: Done. Resolved in commit e9c61052

src/hydro/hydro_cuda.cu
bcaddy: I looked into this. While I'm not 100% positive I'm pretty it's fine. Similar averaging is done in other places

src/integrators/VL_3D_cuda.cu
bcaddy: The two different updates should be entirely independent and could be run in separate streams

bcaddy:
I think I figured out the issue with builds randomly failing. The error ended up being too many resources consumed. I reduced the number of jobs that `Make` launched and it seems to have fixed it

src/utils/gpu.hpp
```diff
@@ -53,6 +53,7 @@ static constexpr int maxWarpsPerBlock = 1024/WARPSIZE;
 #define cudaMemcpy hipMemcpy
 #define cudaMemcpyAsync hipMemcpyAsync
 #define cudaMemcpyPeer hipMemcpyPeer
+#define cudaMemcpyPeer hipMemcpyPeer
```
evaneschneider, line 56: Duplication

src/grid/grid3D.cpp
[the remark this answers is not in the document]
bcaddy: After thinking about this some more I decided to remove it from the Grid3D class entirely. It's not a grid thing, doesn't need any class machinery that simply passing it `Grid3D const &G` doesn't provide, and now all the MHD stuff is in the `mhd` namespace. See commit 7ffad07

src/hydro/hydro_cuda.cu
bcaddy: I believe this is resolved so I'm marking it accordingly. Feel free to reopen if that's not the case

src/system_tests/input_files/tMHDSYSTEMSodShockTubeParameterizedMpi_CorrectInputExpectCorrectOutput.txt
```diff
@@ -0,0 +1,57 @@
+#
```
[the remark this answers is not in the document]
bcaddy: I changed the comment so it's marked as 3D now.

This is temporary. The current hydro sod shock tube test doesn't work with MHD since MHD doesn't have PPMP implemented yet, only PCM. I looked at generalizing the hydro test but it was more work than it was worth so I made a temporary test for the MHD version of the sod shock tube.

src/utils/gpu.hpp
bcaddy: Fixed. Commit will be up shortly.

src/reconstruction/pcm_cuda.cu
```diff
@@ -269,224 +270,179 @@ __global__ void PCM_Reconstruction_3D(Real *dev_conserved,
...
-      scalar[i] = dev_conserved[(5+i)*n_cells + id];
-    }
-    #endif
+      for (int i=0; i<NSCALARS; i++)
+      {
+        scalar[i] = dev_conserved[(5+i)*n_cells + id];
+      }
+    #endif  //SCALAR
+    #ifdef  MHD
+      Real cellCenteredBx, cellCenteredBy, cellCenteredBz;
```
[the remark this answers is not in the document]
bcaddy: Because those are the cell centered fields and they're computed with the `mhd::utils::cellCenteredMagneticFields` function. That function was written before we had C++17 support and so it doesn't use structured binding so the centered magnetic fields had to be passed it and set. I'm going to try it with a structured binding return instead

src/mhd/magnetic_update.cu
```diff
@@ -0,0 +1,81 @@
...
+                                             Real const dy,
+                                             Real const dz)
+    {
+        // get a thread index
+        int const blockId  = blockIdx.x + blockIdx.y*gridDim.x;
+        int const threadId = threadIdx.x + blockId * blockDim.x;
+        int xid, yid, zid;
+        cuda_utilities::compute3DIndices(threadId, nx, ny, xid, yid, zid);
+
+        // Thread guard to avoid overrun and to skip ghost cells that cannot be
```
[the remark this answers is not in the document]
bcaddy: The `-1` should be `-2` in the thread guard. Fixed in 05f086ee

src/mhd/magnetic_divergence.cu
```diff
@@ -0,0 +1,113 @@
...
+                                           int const nx,
+                                           int const ny,
+                                           int const nz,
+                                           int const n_cells)
+    {
+        // First let's create some variables we'll need.
+        cuda_utilities::AutomaticLaunchParams static const launchParams(mhd::calculateMagneticDivergence);
+        cuda_utilities::DeviceVector<Real> static dev_maxDivergence(1);
+
+        // Set the device side inverse time step to the smallest possible double
```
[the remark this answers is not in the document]
bcaddy: Fixed in 1c50a705

src/mhd/ct_electric_fields.cu
```diff
@@ -0,0 +1,341 @@
+/*!
+ * \file ct_electric_fields.cu
+ * \author Robert 'Bob' Caddy (rvc@pitt.edu)
+ * \brief Contains implementation for the CT electric fields code
```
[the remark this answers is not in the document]
bcaddy: Good point. I've been lax about that but I'll add it.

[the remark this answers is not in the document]
bcaddy: Done. Commit 0be3a500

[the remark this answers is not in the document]
bcaddy: And commit ef4f4035

PR #222, opened 2023-01-11
use starburst 99 SN rate info

ojwg:
Starburst 99 generated file included in src code.  This will be placed elsewhere when refactoring feedback code out from the `src/particles` directory.


src/particles/feedback_CIC_gpu.cu
```diff
@@ -21,7 +25,8 @@
 namespace supernova {
   curandStateMRG32k3a_t*  randStates;
   part_int_t n_states;
-  Real t_buff, dt_buff;
+  Real *d_snr, snr_dt, t_sn_start, t_sn_end;
```
bcaddy, line 28: We decided that we would use the `dev_` prefix for device pointers.

src/particles/feedback_CIC_gpu.cu
```diff
@@ -43,27 +48,81 @@ __device__ double atomicMax(double* address, double val)
...
+ * -# Initialize the cuRAND state, which is analogous to the concept of generators in CPU code.
+ *    The state object maintains configuration and status the cuRAND context for each thread on the GPU.
+ *    Initialize more than the number of local particles since the latter will change through MPI transfers.
  *
- * @param n_local
+ * @param P pointer to parameters struct. Passes in starburst 99 filename and random number gen seed.
+ * @param n_local  number of local particles on the GPU
  * @param allocation_factor
  */
 void supernova::initState(struct parameters *P, part_int_t n_local, Real allocation_factor) {
```
bcaddy, line 65: I'm glad we're starting to use more namespaces. It makes generic names like this more clear

src/particles/feedback_CIC_gpu.cu
```diff
@@ -73,6 +132,15 @@ void supernova::initState(struct parameters *P, part_int_t n_local, Real allocat
 }
 
 
+__device__ Real GetSNRate(Real t, Real* d_snr, Real snr_dt, Real t_start, Real t_end) {
```
bcaddy, line 135: should this be in the `supernova` namespace as well?

src/particles/feedback_CIC_gpu.cu
bcaddy, line 28: What does the `t_` prefix mean?

bcaddy, line 28: Can `snr_dt` be replaced with a `DeviceVector` and reduction like in the hydro time step calculations

ojwg: I'll make the change.

src/particles/feedback_CIC_gpu.cu
ojwg: I don't expect this function to be used outside this file, so didn't think to put it into the namespace.

bcaddy: Ok, then you might be better off putting it in an anonymous namespace

src/particles/feedback_CIC_gpu.cu
ojwg: time

ojwg: snr_dt is just a Real number

bcaddy: IMO it might be worth spelling out since `t` can be so many different things

PR #237, opened 2023-02-02
Refactor HLLD solver with structs

bcaddy:
This makes passing all the variables around much simplier, makes the solver much more readable, and enables us to put pretty much everything into testable functions.

### DIFF
--- a/src/riemann_solvers/hlld_cuda.cu
+++ b/src/riemann_solvers/hlld_cuda.cu
@@ -43,579 +43,457 @@ __global__ void Calculate_HLLD_Fluxes_CUDA(Real *dev_bounds_L,
   int xid, yid, zid;
   cuda_utilities::compute3DIndices(threadId, nx, ny, xid, yid, zid);
 
+  // Thread guard to avoid overrun
+  if (xid >= nx and yid >= ny and zid >= nz) return;
+
   // Number of cells
   int n_cells = nx * ny * nz;
 
   // Offsets & indices
   int o1, o2, o3;
   if (direction == 0) {
-    o1 = 1;
-    o2 = 2;
-    o3 = 3;
+    o1 = grid_enum::momentum_x;
+    o2 = grid_enum::momentum_y;
+    o3 = grid_enum::momentum_z;
   }
   if (direction == 1) {
-    o1 = 2;
-    o2 = 3;
-    o3 = 1;
+    o1 = grid_enum::momentum_y;
+    o2 = grid_enum::momentum_z;
+    o3 = grid_enum::momentum_x;
   }
   if (direction == 2) {
-    o1 = 3;
-    o2 = 1;
-    o3 = 2;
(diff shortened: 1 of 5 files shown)


bcaddy:
I found a weird memory access issue with this. Please disregard this PR until I get the issue figured out

src/riemann_solvers/hlld_cuda.cu
```diff
@@ -43,579 +43,457 @@ __global__ void Calculate_HLLD_Fluxes_CUDA(Real *dev_bounds_L,
   int xid, yid, zid;
   cuda_utilities::compute3DIndices(threadId, nx, ny, xid, yid, zid);
 
+  // Thread guard to avoid overrun
+  if (xid >= nx and yid >= ny and zid >= nz) return;
```
alwinm, line 47: should these be "or"?

bcaddy: Yep, converting the "and" to an "or" fixed it.

I agree that `zid >= nz` is an equivalent and faster check but IMO it's much less clear about what the thread guard is trying to do with minimal improvement in performance.

evaneschneider: I agree with @bcaddy - it is also more robust to possible changes in the order of indexing (not that I have any reason to expect that, but you never know).

bcaddy: A function for thread guards would solve this and allow for easy changes to indexing and performance tuning. But IMO we're far from the level of efficiency where tuning our thread guards will matter.

bcaddy:
The memory issue has been fixed in commit 334aa64 and so this is ready to go again

src/riemann_solvers/hlld_cuda.cu
alwinm: I think if we have a standard form of thread guard, my preference then would be 

if (threadID >= nx*ny*nz) return;

and/or

if (threadID >= n_cells) return;

before computing the 3D indices. I think this is both clear and efficient.

bcaddy: Not all kernels can operate on all cells though. Some need a buffer of X cells around them in which case no single check will work. So I don't think we can have a standard thread guard and listing out the exact range of each kernel with the current kind is the closest to standard we can do

alwinm: Good point. I still prefer my suggestion whenever possible (for me personally I like that it avoids boolean logic and/or bugs, and having to read and mentally parse boolean logic) but we can agree to disagree. 

For me personally, I prefer seeing 

```
(tid >= n_cells) return;
// vs.
(tid >= nx*ny*nz) return;
// vs. 
(tid_x >= nx || tid_y >= ny || tid_z >= nz ) return;
```

But anyway. We can agree to disagree, just wanted to share my pov.

PR #248, opened 2023-02-07
Improve restart quality

alwinm:
A few changes improve the consistency of outputs between running uninterrupted from t=0 vs interrupting and restarting a run. 

- During any given timestep, dti is calculated -- the simulation now computes dti consistently before the first loop, whether for init=Read_Grid or for a normal run. 
- For Gravity, the potential and timestep of the previous timestep are needed to extrapolate a half-timestep value. Since this requires saving and reading more data during restarts, I've wrapped this in the ```GRAVITY_RESTART``` flag. 

Other minor changes:
- Write_HDF5_Dataset replaces HDF5_Dataset, I finally pay penance for my sins
- OMP_NUM_THREADS for frontier set to 7
- kernel_printf function provided for easily identifying and disabling printfs within kernels... The utility of this may be questionable. 
- Extra H5open and H5close deleted
- Variety of ```feedback_CIC_gpu.cu``` changes that I've made @ojwg aware of. Orlando has access to a branch where I've made further refactoring changes but I'm not pushing that to dev for now.


builds/make.type.disk
```diff
@@ -36,12 +36,11 @@ DFLAGS    += -DTEMPERATURE_FLOOR
...
 #DFLAGS    += -DCLOUDY_COOL
 DFLAGS    += -DDE
-#DFLAGS    += -DCPU_TIME
+DFLAGS    += -DCPU_TIME
 DFLAGS    += -DAVERAGE_SLOW_CELLS
 DFLAGS    += -DHYDRO_GPU
 
-OUTPUT    ?=  -DOUTPUT -DHDF5
-# -DSLICES
+OUTPUT    ?=  -DOUTPUT -DHDF5 -DSLICES -DPROJECTION
```
bcaddy, line 43: Should we be outputting slices and projections by default?

src/global/global_cuda.h
```diff
@@ -116,6 +116,20 @@ __device__ double atomicAdd(double *address, double val)
 }
     #endif
 
+  // This helper function exists to easily enable/disable printfs inside kernels
+  // And makes it easier to find printfs inside kernels
+
+    #ifndef DISABLE_KERNEL_PRINTF
```
bcaddy, line 122: Won't this lead to issues in `printf`s that aren't in kernels? Or is the compiler smart enough to overload it?

I'm not sure that this has utility. `printf`s in kernels tend to produce a TON of output and really slow down a simulation so they're pretty hard to accidentally leave enabled and I think we should discourage people from using them.

src/gravity/gravity_restart.cpp
```diff
@@ -0,0 +1,100 @@
...
+void Gravity_Restart_Filename(char* filename, char* dirname, int nfile)
+{
+#ifdef MPI_CHOLLA
+  sprintf(filename, "%s%d_gravity.h5.%d", dirname, nfile, procID);
+#else
+  sprintf(filename, "%s%d_gravity.h5", dirname, nfile);
+#endif
+}
+
+#if defined(GRAVITY) && defined(GRAVITY_RESTART) && defined(HDF5)
```
bcaddy, line 29: IMO the code within `GRAVITY_RESTART` should just be part of the `GRAVITY` flag. As a user you would expect gravity to run with all the stuff required for a restart and any serious simulation is going to have to have it on.

src/gravity/gravity_restart.cpp
```diff
@@ -0,0 +1,100 @@
...
+  H5Sclose(dataspace_id);
+
+  H5Fclose(file_id);
+
+  H5close();
+}
+
+#elif defined(GRAVITY)
+// Do nothing
+void Grav3D::Read_Restart_HDF5(struct parameters* P, int nfile) {}
```
bcaddy, line 97: Maybe have these print a warning message instead of doing nothing? Like "gravity restarting not enabled. Skipping writing gravity restart files".

src/global/global_cuda.h
bcaddy: I just saw all the `printf`s in the feedback code. See my comment there about them

src/particles/feedback_CIC_gpu.cu
```diff
@@ -254,14 +275,15 @@ __global__ void Cluster_Feedback_Kernel(
...
+    // kernel_printf("(%d): pos:(%.4e, %.4e, %.4e)\n", gtid, pos_x, pos_y,
+    // pos_z); kernel_printf("(%d): MIN:(%.4e, %.4e, %.4e)\n", gtid, xMin, yMin,
+    // xMin);
 
     bool in_local = (pos_x >= xMin && pos_x < xMax) &&
                     (pos_y >= yMin && pos_y < yMax) &&
                     (pos_z >= zMin && pos_z < zMax);
     if (!in_local) {
-      printf(
+      kernel_printf(
```
bcaddy, line 286: What's the purpose of all the `printf`s?

builds/make.type.debug_disk
```diff
@@ -0,0 +1,56 @@
+DFLAGS += -DDISABLE_KERNEL_PRINTF
```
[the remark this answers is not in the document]
evaneschneider: I agree -- I am strongly trying to _discourage_ the use of new macros unless absolutely necessary. So if this is just a debugging tool, I think it should be commented out for the production branch.

src/gravity/gravity_restart.cpp
evaneschneider: Again, I agree with Bob - I don't like adding new macros that a user has to be aware of and turn on in order to get what one assumes would be the expected behavior.

src/global/global_cuda.h
alwinm: > Won't this lead to issues in printfs that aren't in kernels? Or is the compiler smart enough to overload it?

No, ```#define kernel_printf printf``` replaces ```kernel_printf``` with ```printf``` as default behavior, and replaces it with a do-nothing function to disable printfs. It does not touch printfs elsewhere.

src/gravity/gravity_restart.cpp
alwinm: I'm supportive of this. I added the flag so as to not surprise pre-existing users because previously Cholla did not do this. Happy to remove the flag.

src/particles/feedback_CIC_gpu.cu
alwinm: Not my purview. The scope of the changes here was only to rename some printfs within kernels with an easy to way to find them all.

bcaddy: Fair enough.

builds/make.type.debug_disk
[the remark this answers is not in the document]
alwinm: Removed

src/gravity/gravity_restart.cpp
alwinm: Done

builds/make.type.disk
alwinm: My opinion is that slices and projections are cheap and more frequently needed than not for the disk problem, so defaulting to output is better. Especially to minimize the number of manual edits and changes needed to run production.

evaneschneider: Yeah, I think it's fine to have these on since everyone who runs disk problems uses them.

PR #251, opened 2023-02-09
Refactor write HDF5 attributes for int and double and  increase clang-format width to 120

alwinm:


alwinm:
After hearing positively from Bob, Helena, and Orlando I've converted the draft into a PR.

PR #252, opened 2023-02-14
Remove outflow boundary condition diode for MHD

bcaddy:
# Disable diode outflow boundaries for MHD

Since MHD needs a larger stencil the diode in the outflow BC's was causing issues. After discussion with @evaneschneider the diode has been disabled for MHD builds.

# Clang Tidy

Minor tweaks to fix clang-tidy's `readability-qualified-auto` and `clang-analyzer-optin.cplusplus.UninitializedObject` checks.

The readablity check was fixed by commenting out some currently used code, it's not deleted since it will be used again when MHD is finished; at which time the actual error will be fixed.

The clang-analyzer check was fixed by initializing some member variables in the OneTime class. @alwinm, I'd like you to take a look at this and make sure I haven't broken anything.

# Other

- Switched more indexing over to using grid_enums and added enums for the CT electric fields.
- Cleaned up the arguments for the HLLD solver, removed unused ones, etc[
- Fix issues with Dai & Woodward shock tube initial conditions
- Up the size of all MHD shock tubes from 32^3 to 64^3

### DIFF
--- a/examples/3D/Dai_and_Woodward.txt
+++ b/examples/3D/Dai_and_Woodward.txt
@@ -7,11 +7,11 @@
 
 ################################################
 # number of grid cells in the x dimension
-nx=32
+nx=64
 # number of grid cells in the y dimension
-ny=32
+ny=64
 # number of grid cells in the z dimension
-nz=32
+nz=64
 # final output time
 tout=0.2
 # time interval for output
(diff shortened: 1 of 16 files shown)


alwinm:
> initializing some member variables in the OneTime class.

This looks great to me! I think all those things were left uninitialized because they get set/overwritten later anyway.

PR #265, opened 2023-03-10
Update Radiative Transfer for MPI boundaries

evaneschneider:
This PR incorporates:
-  all the changes from the development branch from the last ~8 months
- support for log time stepping
- the basic framework for running radiative transfer simulations with MPI

Only the x-boundary transfer seems to be working correctly at the moment, so this version can currently only run correctly with a maximum of 2 GPUs.


evaneschneider:
The HIP builds are broken but the CUDA builds pass except for clang tidy.

PR #270, opened 2023-03-21
Fix MHD I/O & Add Restart Test

bcaddy:
## Restart Test

Added a test for MHD and hydro restarts. It doesn't work with gravity, particles, etc but would be pretty easy to extend to those or write new tests for them based off of this test.

## MHD I/O

The MHD I/O was outputting an extra cell in every direction when it should only output an extra cell in the direction of the magnetic field; i.e. the Bx dataset should be (nx+1, ny, nz) not (nx+1, ny+1, nz+1). This should fix that issue and clean up MHD I/O a bit in general.

## Replaced all custom `PI` macros with the built in `M_PI`

### DIFF
--- a/python_scripts/cat_dset_3D.py
+++ b/python_scripts/cat_dset_3D.py
@@ -51,10 +51,9 @@ for n in range(ns, ne+1):
       except KeyError:
         print('No Dual energy data present');
       try:
-        [nx_mag, ny_mag, nz_mag] = head['magnetic_field_dims']
-        bx = fileout.create_dataset("magnetic_x", (nx_mag, ny_mag, nz_mag), chunks=True, dtype=filein['magnetic_x'].dtype)
-        by = fileout.create_dataset("magnetic_y", (nx_mag, ny_mag, nz_mag), chunks=True, dtype=filein['magnetic_y'].dtype)
-        bz = fileout.create_dataset("magnetic_z", (nx_mag, ny_mag, nz_mag), chunks=True, dtype=filein['magnetic_z'].dtype)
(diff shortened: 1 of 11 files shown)


bcaddy:
I'd like @alwinm to take a look at the changes to the I/O routines in `io_gpu.cu`. I borrowed some ideas from my CT kernel for the indexing and did a little bit of refactoring, mostly to reduce code reuse and increase `const`ness.

bcaddy:
This is pretty much ready. Once I get @alwinm's go ahead I just need to update the test data that is impacted by this and it will be good to go.

alwinm (commented):
Looks good to me

@bcaddy I think yeeting the PI macro in favor of M_PI could be done uniformly throughout the code. I like that it's easier to grep, though I might actually prefer something like MACRO_PI or PI_MACRO since M_PI is a bit misleading in an MPI code haha. Would you want to? Does not need to be in this PR. 

There's one completely unused here:
https://github.com/cholla-hydro/cholla/blob/dev/src/system_tests/cooling_system_tests.cpp

The other occurrences are in initial conditions. 
https://github.com/cholla-hydro/cholla/blob/dc2b21c7ef2c91b0a57327c79ade497db23dcc34/src/grid/initial_conditions.cpp#L278

bcaddy:
Thanks for checking it out.

I agree that `PI` should be replaced with `M_PI`. I agree that the naming is unfortunate in an MPI code but it is the standard so I think we should stick with it until we switch to C++20 which has a much clearer implementation of [mathematical constants](https://en.cppreference.com/w/cpp/numeric/constants).

bcaddy:
@alwinm, I got bored while the tests were running and replaced all the instances of `PI` with `M_PI`

bcaddy:
@evaneschneider, once the tests pass this is good to merge

PR #274, opened 2023-03-22
Add MHD blast wave test & initial conditions

bcaddy:
## Summary 

- Example parameter file in examples/3D/mhd_blast.txt
- New parameter `P_blast` parameter to indicate the pressure in the over pressure zone of the blast
- New initial conditions type `MHD_Spherical_Blast` and an initializing function of the same name
- New MHD system test for the MHD blast wave
- New check in `Check_Configuration` to check that gamma is greater than one. I accidentally forgot to set gamma when setting up this test originally so I added a check to make sure that doesn't happen again.

### DIFF
--- a/src/global/global.cpp
+++ b/src/global/global.cpp
@@ -361,6 +361,8 @@ void parse_param(char *name, char *value, struct parameters *parms)
     parms->polarization = atof(value);
   } else if (strcmp(name, "R") == 0) {
     parms->R = atof(value);
+  } else if (strcmp(name, "P_blast") == 0) {
+    parms->P_blast = atof(value);
 #ifdef PARTICLES
   } else if (strcmp(name, "prng_seed") == 0) {
     parms->prng_seed = atoi(value);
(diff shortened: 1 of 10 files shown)


src/global/global.cpp
```diff
@@ -359,6 +359,10 @@ void parse_param(char *name, char *value, struct parameters *parms)
     parms->yaw = atof(value);
   } else if (strcmp(name, "polarization") == 0) {
     parms->polarization = atof(value);
+  } else if (strcmp(name, "R") == 0) {
```
alwinm, line 362: You might consider renaming R to ```R_blast``` similarly to how you've named ```P_blast``` which I think is quite descriptive. 

Although I am a proponent of the single letter variables, you can also consider replacing "R" and "P" with "radius" and "pressure"

bcaddy: Ok, I've changed `R` to `radius`. I think I'm going to leave `P_blast` alone because I don't want to change `P` in every single parameter file and I want it to be clear that `P` and `P_blast` are the same thing, just with slightly different uses.

alwinm (approved):
lgtm

evaneschneider (approved):

PR #276, opened 2023-03-25
RT MPI debug

alwinm:
Sorry about the very ugly diff in RT_kernels.cu resulting from running dos2unix in a regrettable effort to remove ^M endings. The main difference is a number of ```nz``` were converted to ```ny```

After these bugfixes, the iliev1.txt test works and is consistent between 1, 2, 4, and 8 GPUs.


alwinm:
The bit in ```mpi/mpi_routines.cpp``` can probably be removed or replaced with some kind of comment. It currently does nothing but I left it in for now... if nfreq gets larger then it would need to be adjusted appropriately.

PR #277, opened 2023-03-27
Miscellaneous Post MHD Tweaks/Fixes

bcaddy:
# Summary

This is a handful of small tweaks and fixes I've been putting off while I worked on debugging MHD. Most are minor but I would like @evaneschneider or @bvillasen to take a look at my change to the grid temperature function. I think there might be other places that should be using MHD energy in computations but I don't know where

## List of Changes

- Have `make tidy` print out where the results are stored. This isn't obvious so now it tells you
- Replace straight sum in `systemTest::SystemTestRunner::runL1ErrorTest` with a Kahan sum to avoid loss of precision.
- Fix a minor bug in hydro restart test
- Replace file manipulation methods in `systemTest::SystemTestRunner` with `std::filesystem` calls

### DIFF
--- a/src/chemistry_gpu/chemistry_functions.cpp
+++ b/src/chemistry_gpu/chemistry_functions.cpp
@@ -254,7 +254,10 @@ void Grid3D::Compute_Gas_Temperature(Real *temperature, bool convert_cosmo_units
   #ifdef DE
         GE = C.GasEnergy[id];
   #else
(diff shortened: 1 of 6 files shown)


src/chemistry_gpu/chemistry_functions.cpp
evaneschneider, line 1: Looks good to me!

evaneschneider: Actually this is a bug - the relevant header files are not included so chemistry builds break. Will add an issue.

PR #279, opened 2023-03-31
Dev clang tidy single failure checks

bcaddy:
## modernize-use-emplace, hicpp-use-emplace e24b94d

These checks are for cases when the programmer is using `push_back` when it would be easier, simpler, and more performant to use `emplace_back`.

## cert-str34-c e666f85

This check is an alias to bugprone-signed-char-misuse which has already been fixed

## cert-dcl50-cpp af8f683

This check prohibit C-style variadic functions (the ones with ...). We only have one, `chprintf`, and since that's basically an MPI safe wrapper around `printf` I don't think we should change it. Instead I marked it as ok using the NOLINT comment and enabled the check.

## clang-analyzer-core.uninitialized.Assign b0c1be0 and 82fdc28

Potentially uninitialized variables used. They'sr not uninitialized in practice but I went ahead and initialized them anyway. 

## clang-analyzer-valist.Uninitialized f16e613

This only has one error and it's in `chprintf`. The code appears to be correct as is so I'm marking the offending line as NOLINT for this check.

## hicpp-multiway-paths-covered 745b3c2

This check checks for missing `else`s in `if` statements and `default` in `switch` statments. We have many cases where there's no need for a fall through case so I've added it to the list of permanently disable checks.


helenarichie:
These all look good!

helenarichie (approved):

evaneschneider (approved):

Issue #241, opened 2023-02-03
Deprecate CUDA macro


evaneschneider:

bcaddy:
Does this mean we should just remove `CUDA` macros as we come across them? Or do it all in one fell swoop?

evaneschneider:
Yes, with the exception of course of any code that controls the CUDA / HIP compile conversion.

bcaddy:
Resolved in #360

PR #282, opened 2023-03-31
Add doxygen comments to dust code

helenarichie:
Updated existing comments in dust_cuda.cu and added Doxygen comments to dust_cuda.h, dust_cuda.cu, and dust_cuda_tests.cpp. I'm unsure I used `\param[in]` and `\param[out]` appropriately in the function doc strings dust_cuda.h. I think that `\param[out]` should be used for an input parameter that I'm modifying, but not returning in the function, but in the case of `dev_conserved` I don't know if I should specify the field that I'm modifying or something like that. If anyone has thoughts on this, please let me know!


src/dust/dust_cuda.cu
```diff
@@ -1,12 +1,23 @@
...
+ * McKinnon et al. (2017) model of dust sputtering, which depends on the cell's gas density and temperature.
+ */
+
 #ifdef CUDA
   #ifdef DUST
 
+    // STL includes
     #include <stdio.h>
 
+    // External includes
```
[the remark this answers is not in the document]
bcaddy: I concur. Those are all part of the C++ STL (Standard Template Library, usually just called the "standard library")

bcaddy:
Here's what the `\param` in/out stuff means

- `\param[in]` is for variables that are input only and will not be modified in the function. They should also be declared as `const` or pass by value in the function definition to enforce no unexpected changes
- `\param[out]` is for variables that are output only. Their initial values will not be used and they should be passed in by reference or they should be a pointer type. Generally in this case you should just be returning the value from the function with a `return` statement; possibly with structured binding if needed.
- `\param[in, out]' is for variables that are both input and output. Their initial values will be used and the will also contain output; they should be passed in by reference or they should be a pointer type. `dev_conserved` is almost always either `[in]` or `[in,out]`.

src/dust/dust_cuda.h
```diff
@@ -9,13 +16,58 @@
...
 
+/*!
+ * \brief Launch the dust kernel.
+ *
+ * \param[in] dev_conserved The device conserved variable array.
+ * \param[in] nx Number of cells in the x-direction
+ * \param[in] ny Number of cells in the y-direction
+ * \param[in] nz Number of cells in the z-direction
+ * \param[in] n_ghost Number of ghost cells
+ * \param[in] n_fields
```
bcaddy, line 27: This needs a description

src/dust/dust_cuda.h
```diff
@@ -9,13 +16,58 @@
...
+ * \param[in] dt Simulation timestep
+ * \param[in] gamma Specific heat ratio
+ *
+ * \return None
```
bcaddy, line 31: This line isn't needed since the function is `void`

src/dust/dust_cuda.h
```diff
@@ -9,13 +16,58 @@
...
+/*!
+ * \brief Compute the change in dust density for a cell and update its value in dev_conserved.
+ *
+ * \param[out] dev_conserved The device conserved variable array. The dust field is updated in this function. If dual
+ * energy is turned on, then the dual energy field is updated, as well.
+ * \param[in] nx Number of cells in the x-direction
+ * \param[in] ny Number of cells in the y-direction
+ * \param[in] nz Number of cells in the z-direction
+ * \param[in] n_ghost Number of ghost cells
+ * \param[in] n_fields
```
bcaddy, line 44: description

src/dust/dust_cuda.h
```diff
@@ -9,13 +16,58 @@
...
+ * \param[in] dt Simulation timestep
+ * \param[in] gamma Specific heat ratio
+ *
+ * \return None
```
bcaddy, line 48: unneeded

src/dust/dust_cuda.cu
```diff
@@ -1,12 +1,23 @@
...
     #include <cstdio>
     #include <fstream>
     #include <vector>
 
+    // Local includes
```
bcaddy, line 20: This isn't strictly part of the PR. I think that the includes for `dust_cuda.h` should include the full path: `../dust/dust_cuda.h` to be consistent with all the other includes in the code. It also enables easy automated changes if a file gets moved

bcaddy (changes requested):
Overall looks great. A few minor things that need addressed but nothing major.

bcaddy (approved):
Looks good, thank you

src/dust/dust_cuda.cu
```diff
@@ -1,20 +1,30 @@
+/*!
+ * \file dust_cuda.cu
+ * \author Helena Richie (helenarichie@gmail.com)
+ * \brief Contains code that updates the dust density scalar field. The dust_kernel function determines the rate of
+ * change of dust density, which is controlled by the sputtering timescale. The sputtering timescale is from the
+ * McKinnon et al. (2017) model of dust sputtering, which depends on the cell's gas density and temperature.
+ */
+
 #ifdef CUDA
```
evaneschneider, line 9: I am actually trying to deprecate the CUDA macro (see Issue #241) since in practice Cholla can only be run in GPU mode these days, so let's go ahead and get rid of it in new code.

src/dust/dust_cuda.h
```diff
@@ -1,3 +1,10 @@
+/*!
+ * \file dust_cuda.h
+ * \author Helena Richie (helenarichie@pitt.edu)
+ * \brief Contains the declaration for the kernel that updates the dust density scalar in dev_conserved.
+ *
+ */
+
 #ifdef CUDA
```
evaneschneider, line 8: Same here.

evaneschneider:
It looks like this failed on one of Bob's new MHD tests - can you try updating the submodule again?

evaneschneider:
(It's probably my fault for auto-merging in the most recent PR - I guess that doesn't update the submodule.)

PR #284, opened 2023-04-04
Add Orszag-Tang vortex test & initial conditions

bcaddy:
## Summary 

- Example parameter file in `examples/3D/orszag_tang_vortex.txt`
- New initial conditions type `Orszag_Tang_Vortex` and an initializing function of the same name
- New MHD system test for the Orszag-Tang vortex

### DIFF
--- a/src/system_tests/mhd_system_tests.cpp
+++ b/src/system_tests/mhd_system_tests.cpp
@@ -613,6 +613,7 @@ TEST_P(tMHDSYSTEMParameterizedMpi, MhdBlastWaveCorrectInputExpectCorrectOutput)
 TEST_P(tMHDSYSTEMParameterizedMpi, OrszagTangVortexCorrectInputExpectCorrectOutput)
 {
   test_runner.numMpiRanks = GetParam();
+  test_runner.setFixedEpsilon(8.E-4);
   test_runner.runTest();
 }
 /// @}
--- a/src/system_tests/mhd_system_tests.cpp
+++ b/src/system_tests/mhd_system_tests.cpp
@@ -608,6 +608,13 @@ TEST_P(tMHDSYSTEMParameterizedMpi, MhdBlastWaveCorrectInputExpectCorrectOutput)
   test_runner.numMpiRanks = GetParam();
   test_runner.runTest();
 }
+
+/// Test the Orszag-Tang Vortex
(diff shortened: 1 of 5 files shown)


helenarichie (commented):
Just to clarify, was this loosening only supposed to be temporary? I noticed that in the final version of this file, you removed the `test_runner.setFixedEpsilon(8.E-4);` line, so I was wondering if that meant you expect the variation for this test between machines to be small enough that you don't need it ultimately.

src/system_tests/mhd_system_tests.cpp
```diff
@@ -613,6 +613,7 @@ TEST_P(tMHDSYSTEMParameterizedMpi, MhdBlastWaveCorrectInputExpectCorrectOutput)
 TEST_P(tMHDSYSTEMParameterizedMpi, OrszagTangVortexCorrectInputExpectCorrectOutput)
 {
   test_runner.numMpiRanks = GetParam();
+  test_runner.setFixedEpsilon(8.E-4);
```
helenarichie, line 616: Just to clarify, was this loosening only supposed to be temporary? I noticed that in the final version of this file, you removed the `test_runner.setFixedEpsilon(8.E-4);` line, so I was wondering if that meant you expect the variation for this test between machines to be small enough that you don't need it ultimately.

bcaddy: I'm not sure what you mean, that is in the final version of the file. By default the fixed epsilon, i.e. allowed absolute error, is 1E-14, for this test I had issues with reproducibility across C-3PO and H2P despite the fact that the results were qualitatively identical. Since this particular test is qualitative, rather that having a specific quantity we can test, I increased the allowed absolute error to get it to pass. It's not a great solution but the test is mostly to track any issues that might show up in a complex flow rather than prove correctness. It's more a regression test than a correctness test.

helenarichie: You're right, that's my bad--I was looking at the commits in the wrong order... Sorry about that!

helenarichie (approved):

src/system_tests/mhd_system_tests.cpp
bcaddy: No worries. Thank you for checking!

evaneschneider (approved):

PR #287, opened 2023-04-11
HDF5 Refactor + Read_Cat + ScopedTimer

alwinm:
Major breaking change: the scalar loop in hdf5 output from here on will be invalid, all advected scalars should have their own names (rather than scalar1, scalar2, etc.), in order to avoid awful nested ifdef contraptions. 

Refactor HDF5 input/output for:
- Moved 1/2/3D if statements so that we do not need 3 copies of code
- Chemistry/Gravity/Scalars refactored along similar lines as hydro variables
- 1 line for each field input, 1 line for each field output. 

Add ```init=Read_Grid_Cat``` function for restarting from a concatenated file (WARNING: this is slower on most systems, though performance of reading only occurs once, so currently not worth worrying about. This function should be updated by anyone actually using it to include desired fields)

Add ```ScopedTimer``` timing utility as a one-liner for timing a function. (See ```Grid3D::Read_Grid_Cat``` for example use). Does "nothing" without CPU_TIME.


alwinm:
In this rendition since some of the Chemistry fields are host/cpu-only, I left all of them on the cpu-side. But if desired I believe I could gpu-accelerate all the Grid.Conserved chemistry quantities.

To possibly remove needing to cudamemcpy the grid during output steps, we would need to port all the projection/slice/rotated projection code. Which is likely worth doing in a future PR.

alwinm:
> > Major breaking change: Scalar loop from here on will be invalid, all advected scalars should have their own names (rather than scalar1, scalar2, etc.), in order to avoid awful nested ifdef contraptions.
> 
> Could you expand on this a bit? We have a **lot** of loops over scalars in pretty much every kernel.

Good point. I misspoke. Scalar loop in **hdf5 output** will not be valid. Everywhere else, no change.

bcaddy:
Got it, thank you for clarifying

evaneschneider:
This looks good to me, although I'd like @bcaddy to confirm that the array sizes / buffers for MHD reads and writes are correct. Does this also mean that this commit breaks the scalar loop in e.g. the slice outputs? Or was that already broken because it doesn't reference basic scalar?

alwinm:
> This looks good to me, although I'd like @bcaddy to confirm that the array sizes / buffers for MHD reads and writes are correct. Does this also mean that this commit breaks the scalar loop in e.g. the slice outputs? Or was that already broken because it doesn't reference basic scalar?

That's a good point--slice/projections will behave as before, but I believe we should phase out the scalar loops therein.

evaneschneider:
I will merge this once @bcaddy confirms that the buffers etc. are good for MHD.

bcaddy:
I totally forgot about this, sorry! I'll start on it now.

src/main.cpp
```diff
@@ -237,9 +237,7 @@ int main(int argc, char *argv[])
 #endif  // CPU_TIME
     start_step = get_time();
 
-    // calculate the timestep. Note: this computes the timestep ONLY on the
-    // first loop, on subsequent time steps it just calls the MPI_Allreduce to
-    // determine the global timestep
+    // calculate the timestep by calling MPI_Allreduce
```
bcaddy, line 240: since this really just a wrapper around MPI_Allreduce and doesn't actually compute anything I might change the comment to something like "Perform MPI_Allreduce to determine the timestep"

src/io/io_parallel.cpp
```diff
@@ -0,0 +1,135 @@
...
+  offset[2] = nz_local_start;
+
+  // This is really dims but I name it count because that's what HDF5 names it
+  hsize_t count[3];
+  count[0] = H.nx_real;
+  count[1] = H.ny_real;
+  count[2] = H.nz_real;
+
+  #ifdef MHD
+  Real* dataset_buffer = (Real*)malloc((H.nz_real + 1) * (H.ny_real + 1) * (H.nx_real + 1) * sizeof(Real));
```
bcaddy, line 94: This is technically too big but it's the easiest way to make sure the buffer is big enough.

src/io/io.cpp
```diff
@@ -1221,6 +1237,123 @@ void Write_HDF5_Field_2D_CPU(Header H, hid_t file_id, hid_t dataspace_id, float
...
+void Fill_HDF5_Buffer_From_Grid_CPU(int nx, int ny, int nz, int nx_real, int ny_real, int nz_real, int n_ghost,
+                                    Real *hdf5_buffer, Real *grid_buffer)
+{
+  int i, j, k, id, buf_id;
+  // 3D case
+  if (nx > 1 && ny > 1 && nz > 1) {
+    for (k = 0; k < nz_real; k++) {
+      for (j = 0; j < ny_real; j++) {
+        for (i = 0; i < nx_real; i++) {
+          id                  = (i + n_ghost) + (j + n_ghost) * nx + (k + n_ghost) * nx * ny;
```
bcaddy, line 1250: This should be using the `cuda_utilities::compute1DIndex` function.

src/io/io.cpp
```diff
@@ -2416,6 +2286,64 @@ void Grid3D::Read_Grid_Binary(FILE *fp)
...
+  // And for 2D nz_real is not used.
+  // This protects the magnetic case where ny_real/nz_real += 1
+
+  int i, j, k, id, buf_id;
+  // 3D case
+  if (nx > 1 && ny > 1 && nz > 1) {
+    for (k = 0; k < nz_real; k++) {
+      for (j = 0; j < ny_real; j++) {
+        for (i = 0; i < nx_real; i++) {
+          id              = (i + n_ghost) + (j + n_ghost) * nx + (k + n_ghost) * nx * ny;
```
bcaddy, line 2304: This should be using the cuda_utilities::compute1DIndex function.

bcaddy (approved):
This looks good to me. I don't see any issues with the MHD stuff and it's passing all the tests. I noted a few minor quibbles on non-MHD sections but nothing that is required.

src/io/io_parallel.cpp
evaneschneider: I see.

bcaddy: Something like 
```cpp
size_t const buff_size = std::max({(H.nz_real + 1) * H.ny_real * H.nx_real, H.nz_real * (H.ny_real + 1) * H.nx_real, H.nz_real * H.ny_real * (H.nx_real + 1)});
Real* dataset_buffer = (Real*)malloc(buff_size * sizeof(Real));
```
would get the right size.

src/io/io.cpp
```diff
@@ -1221,6 +1237,123 @@ void Write_HDF5_Field_2D_CPU(Header H, hid_t file_id, hid_t dataspace_id, float
...
+                                    Real *hdf5_buffer, Real *grid_buffer)
+{
+  int i, j, k, id, buf_id;
+  // 3D case
+  if (nx > 1 && ny > 1 && nz > 1) {
+    for (k = 0; k < nz_real; k++) {
+      for (j = 0; j < ny_real; j++) {
+        for (i = 0; i < nx_real; i++) {
+          id                  = (i + n_ghost) + (j + n_ghost) * nx + (k + n_ghost) * nx * ny;
+          buf_id              = k + j * nz_real + i * nz_real * ny_real;
```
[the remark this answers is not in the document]
alwinm: Usually yes but I think in this function in particular leaving it this way makes the re-map of indices more obvious without having to look up the compute1Dindex definition (showing that HDF5 and in-memory are assuming different orderings). Since I see it both ways I'll leave it as is.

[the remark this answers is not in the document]
bcaddy: You can just call it with the arguments in a different order. That should make the difference clear and reduces code duplication and potential for error.

PR #291, opened 2023-04-26
Improve timers and read_grid_cat restart

alwinm:
- Make minor fixes so that Cholla compiles without -DOUTPUT and -DHDF5
- Add Hydro Integrator-specific timer. Note that double-counting time is not an issue since the Total is not computed from any sum. 
- Add CudaSynchronize to timers (necessary for Hydro Integrator)
- Make slight improvement to Read_Cat_Grid to avoid overwriting input and to start at the correct nfile. 
- Make slight improvement to Read_Cat_Grid to correctly set some variables which are not strictly necessary but may cause confusion.

### DIFF
--- a/src/grid/grid3D.cpp
+++ b/src/grid/grid3D.cpp
@@ -426,6 +426,11 @@ Real Grid3D::Update_Grid(void)
   U_floor /= Cosmo.v_0_gas * Cosmo.v_0_gas / Cosmo.current_a / Cosmo.current_a;
 #endif
 
+
+#ifdef CPU_TIME
+  Timer.Hydro_Integrator.Start();
+#endif //CPU_TIME
+
   // Run the hydro integrator on the grid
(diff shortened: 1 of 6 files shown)


bcaddy:
Do the additional `cudaDeviceSync` calls impact performance?

alwinm:
> Do the additional `cudaDeviceSync` calls impact performance?

It's a good question. I haven't looked carefully but it's likely sub-ms or within noise. The more impactful part of the CPU_TIME flag is the MPI collection and synchronization. To squeeze out all the performance juice I would recommend leaving the flag disabled.

alwinm:
Overhead of timers, measured by placing a Start and End adjacent with nothing in between.

```
0.0052   ms
```

bcaddy:
All the tests are passing now. Looks like it was a one off issue

bcaddy:
All the tests are passing now. Looks like it was a one off issue

PR #293, opened 2023-04-28
Fix 7 clang-tidy checks

bcaddy:
## Clang-Tidy Checks

Fixed and enabled the following `clang-tidy` Checks

- performance-unnecessary-value-param check
- performance-inefficient-vector-operation check
- modernize-use-auto and its alias hicpp-use-auto.
- performance-for-range-copy check
- performance-faster-string-find check
- modernize-use-default-member-init check
- modernize-loop-convert tidy check

## Other

- Removed unsafe pointer alias to a DeviceVector pointer in io.cpp
- Append make type to clang-tidy log files


alwinm:
It looks good to me

alwinm (approved):
lgtm

src/grid/boundary_conditions.cpp
```diff
@@ -151,7 +151,7 @@ int Grid3D::Check_Custom_Boundary(int *flags, struct parameters P)
   }
 
   for (int i = 0; i < 6; i++) {
-    if (!((flags[i] >= 0) && (flags[i] <= 5))) {
+    if (flags[i] < 1 or flags[i] > 5) {
```
evaneschneider, line 154: Do you know if these "or" and "not" statements are portable for other compilers?

evaneschneider (approved):

src/grid/boundary_conditions.cpp
bcaddy: They've been part of the C++ standard since C++98 so any standard compliant compiler will support them; see section 2.5 of the [C++98 Standard](https://www.lirmm.fr/~ducour/Doc-objets/ISO+IEC+14882-1998.pdf).

MSVC is historically non-compliant with whatever standards they decide they don't like and so some versions from the mid 2010's don't support these alternative tokens directly but they can be enabled by either [setting the right compiler flags or including the compliance header](https://stackoverflow.com/questions/24414124/why-does-vs-not-define-the-alternative-tokens-for-logical-operators).

PR #300, opened 2023-05-17
PPMC Refactor, New PPMC for Van Leer Integrator, and MHD Support in the New PPMC Kernel

bcaddy:
## PPMC Refactor

Building on the work in PR #288, PPMC has been refactored to use functions

## New PPMC Method

When extended to MHD the old PPMC method was unstable on some problems, especially the Brio & Wu shock tube. The new method is stable and based off of the method currently used in Athena++; it's also faster and should be more stable and accurate for hydro. The old PPMC is still available as the `PPMC_CTU` kernel and is used in the CTU and SIMPLE integrators as the new PPMC method does not support time evolution of the variables. The new PPMC is available as the `PPMC_VL` kernel and is used in the Van Leer integrator.

All tests have been updated for PPMC, new tests have been written for the new functions, etc.

## Other

- Fixed a bug in the characteristic/primitive projection. See [this PR](https://github.com/PrincetonUniversity/athena/pull/497) for more info
- Added a compile time check that only one integrator is enabled

## DRAFT REASON

This is currently a draft until the PLMC PR is merged in. Because it's based on the same branch as the PLMC PR it looks like it has all the changes of both branches. Once the PLMC PR is merged the diff for this PR should only include the PPMC stuff.


bcaddy:
@evaneschneider, FYI, this PR now contains the fixes to the linear wave initial conditions that fix the unexpectedly poor scaling of the fast wave

bcaddy:
FYI, this also fixes a memory leak in `Grid3D::Write_Projection_HDF5` where the memory that `dataset_buffer_dust_xy` and `dataset_buffer_dust_xz` pointed at wasn't freed. Given that that function might be called many times I suggest replacing the various buffers with static STL vectors.

src/reconstruction/plmc_cuda.cu
```diff
@@ -62,11 +62,11 @@ __global__ void PLMC_cuda(Real *dev_conserved, Real *dev_bounds_L, Real *dev_bou
   reconstruction::Primitive const cell_i =
       reconstruction::Load_Data(dev_conserved, xid, yid, zid, nx, ny, n_cells, o1, o2, o3, gamma);
 
-  // cell i-1. The equality checks check the direction and subtract one from the direction
+  // cell i-1. The equality checks check the direction and subtracts one from the direction
```
evaneschneider, line 65: Technically I think this was correct before :)

src/reconstruction/ppmc_cuda.cu
```diff
@@ -1,1277 +1,681 @@
 /*! \file ppmc_cuda.cu
  *  \brief Functions definitions for the ppm kernels, using characteristic
  tracing. Written following Stone et al. 2008. */
-#ifdef CUDA
-  #ifdef PPMC
```
evaneschneider, line 5: Why don't you need the PPMC def?

src/reconstruction/ppmc_cuda.cu
```diff
@@ -1,1277 +1,681 @@
...
-      o1 = 3;
-      o2 = 1;
-      o3 = 2;
-      break;
+  // get a thread ID
+  int const thread_id = threadIdx.x + blockIdx.x * blockDim.x;
+  int xid, yid, zid;
+  cuda_utilities::compute3DIndices(thread_id, nx, ny, xid, yid, zid);
+
+  // Thread guard to prevent overrun
```
evaneschneider, line 30: Can you make this comment clearer? I.e. it used to say something like "Check to make sure we are only operating on real cells" or something like that, now it is a sort of meaningless phrase.

src/reconstruction/ppmc_cuda.cu
```diff
@@ -1,1277 +1,681 @@
...
-    sum_2 = 0;
-    sum_3 = 0;
-    sum_4 = 0;
-    sum_5 = 0;
-      #ifdef DE
-    sum_ge = 0;
-      #endif  // DE
-      #ifdef SCALAR
+  // right
+  del_R = reconstruction::Compute_Slope(cell_ip1, cell_i);
```
evaneschneider, line 147: I find it a little funky that these slopes mostly send the higher index first, but then the VL slope sends the lower index first. Just seems like an easy way to make a bug. I'd have written them to all take the lower index first.

src/reconstruction/ppmc_cuda.h
```diff
@@ -1,21 +1,44 @@
...
+ */
+__global__ void PPMC_CTU(Real *dev_conserved, Real *dev_bounds_L, Real *dev_bounds_R, int nx, int ny, int nz, Real dx,
+                         Real dt, Real gamma, int dir);
 
-    #endif  // PPMC_CUDA_H
-  #endif    // PPMC
-#endif      // CUDA
+/*!
+ * \brief Computes the left and right interface states using PPM with limiting in the characteristic variables. Used for
+ * the VL (Van Leer) integrators
```
evaneschneider, line 30: It would be useful if the description of these two functions also included what is different about the PPM.

evaneschneider (approved):
I made a couple comments, but overall think this is good to go.

src/reconstruction/ppmc_cuda.cu
bcaddy: Because we can compile the kernels just fine without it, they just won't be used anywhere.

src/reconstruction/plmc_cuda.cu
bcaddy: Fixed in 5c315753

src/reconstruction/ppmc_cuda.cu
bcaddy: Good catch, fixed in ef1598b

src/reconstruction/ppmc_cuda.cu
bcaddy: Generally my approach has been that unless an ifdef is absolutely necessary I don't include it (or remove it if I'm refactoring)

evaneschneider: Alright. I guess we don't really live in a world where compile time matters anymore :)

src/reconstruction/ppmc_cuda.h
bcaddy: I added more info in [3b192d3](https://github.com/cholla-hydro/cholla/pull/300/commits/3b192d3f6329764beab332983fcea790e20d91a0). Could you take a look and let me know if that's what you meant?

bcaddy: I think I've addressed it now. It's still a little vague but it's hard to get into details without just reciting an entire paper or the code. It has all the citations and an overall description now anyway.

evaneschneider: Much better, thank you!

PR #305, opened 2023-06-29
Linear Wave Convergence Test

bcaddy:
This PR adds a set of tests that check that the MHD linear waves converge at the right order. This was added because I had an uncaught issue with the linear wave initial conditions that caused first order convergence in the fast wave when we expect second order.

This PR is a draft for now as it's based off of my PPMC branch which has not yet been merged into `dev`. See commit [bf51c49](https://github.com/cholla-hydro/cholla/commit/bf51c49c94598cdfe96959947f1bb36aa81ade60) for the contents of this PR.


src/system_tests/mhd_system_tests.cpp
```diff
@@ -32,15 +33,15 @@ class tMHDSYSTEMLinearWavesParameterizedAngle : public ::testing::TestWithParam<
 
  protected:
   systemTest::SystemTestRunner waveTest;
+  inline static std::unordered_map<std::string, double> n32_l2norms;
```
helenarichie, line 36: You might want to rename this to something like "`low_res_l2norms`". I wasn't sure what the `n32` prefix or the "Check the scaling" comments in the higher-resolution tests meant, so this might clarify those.

helenarichie:
I left one comment about a variable name in the MHD system tests file, but other than that this looks ready to go!

src/system_tests/mhd_system_tests.cpp
bcaddy: Good point. Fixed in commit [166a102](https://github.com/cholla-hydro/cholla/pull/305/commits/166a10230a6eb8d90a24fb13fb702d7ee2114ddb).

PR #306, opened 2023-06-29
Optimize Characteristic Computations in Refactored Reconstructors

bcaddy:
I realized that I was recomputing the eigenvalues for the characteristic projections multiple times for no reason in the PPMC and PLMC reconstruction kernels. This refactor changes it so that the eigenvectors are only computed once and then used in all the projections. Overall this resulted in noticeable performance increases on both a V100 system (C-3PO) and on Frontier. The performance numbers are in this table, all times given are for the entire timestep, not just the PPMC kernel. The test was the Brio & Wu shock tube at 256^3 run with PPMC.

| Test                         | Time (ms/time step) | Registers (scalar/vector on AMD) |
|------------------------------|--------------------:|---------------------------------:|
| C-3PO Initial                |               189.1 |                              157 |
| C-3PO Final                  |               180.4 |                              158 |
| C-3PO Improvement            |                4.6% |                               +1 |
| Frontier/Crusher Initial     |               167.5 |                           54/128 |
| Frontier/Crusher Final       |               144.9 |                           54/128 |
| Frontier/Crusher Improvement |               13.5% |                              0/0 |

With specifying `__launch_bounds__(256)` performance is about 40% improved over baseline on the MI250X

### DIFF
--- a/src/reconstruction/reconstruction.h
+++ b/src/reconstruction/reconstruction.h
@@ -48,7 +48,7 @@ struct Primitive {
 // =====================================================================================================================
 
 // =====================================================================================================================
-struct eigenVecs {
+struct EigenVecs {
   Real magnetosonic_speed_fast, magnetosonic_speed_slow, magnetosonic_speed_fast_squared,
       magnetosonic_speed_slow_squared;
   Real alpha_fast, alpha_slow;
@@ -263,13 +263,13 @@ Primitive __device__ __host__ __inline__ Van_Leer_Slope(Primitive const &left_sl
  * \param[in] sound_speed The sound speed
  * \param[in] sound_speed_squared The sound speed squared
  * \param[in] gamma The adiabatic index
- * \return eigenVecs
+ * \return EigenVecs
  */
 #ifdef MHD
-eigenVecs __device__ __inline__ Compute_Eigenvectors(Primitive const &primitive, Real const &sound_speed,
+EigenVecs __device__ __inline__ Compute_Eigenvectors(Primitive const &primitive, Real const &sound_speed,
                                                      Real const &sound_speed_squared, Real const &gamma)
 {
-  eigenVecs output;
+  EigenVecs output;
   // This is taken from Stone et al. 2008, appendix A. Equation numbers will be quoted as relevant
 
   // Compute wave speeds and their squares
@@ -328,7 +328,7 @@ eigenVecs __device__ __inline__ Compute_Eigenvectors(Primitive const &primitive,
  *
  * \param[in] primitive The primitive variables
  * \param[in] primitive_slope The primitive variables slopes
- * \param[in] eigenVecs The eigenvectors
+ * \param[in] EigenVecs The eigenvectors
  * \param[in] sound_speed The speed of sound
  * \param[in] sound_speed_squared The speed of sound squared
  * \param[in] gamma The adiabatic index
@@ -336,7 +336,7 @@ eigenVecs __device__ __inline__ Compute_Eigenvectors(Primitive const &primitive,
  */
 Characteristic __device__ __inline__ Primitive_To_Characteristic(Primitive const &primitive,
                                                                  Primitive const &primitive_slope,
-                                                                 eigenVecs const &eigen, Real const &sound_speed,
+                                                                 EigenVecs const &eigen, Real const &sound_speed,
                                                                  Real const &sound_speed_squared, Real const &gamma)
 {
   Characteristic output;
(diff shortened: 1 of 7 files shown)


twhite-cray:
The Cuda register use tells me this kernel "wants" more than 128. The AMD kernel is probably limited to 128 vector registers because of the default launch bounds of 1024 threads, which forces an occupancy of 4. You may want to add `__launch_bounds__(256)` to the kernel definition to see if that helps performance by allowing more register use and avoiding register spills.

bcaddy:
Doing that it looks like the only offenders are PLMC_VL, PPMC_VL, and Calculate_Roe_Fluxes_CUDA. We don't use the last one but I'll add the flag to the other two

src/reconstruction/reconstruction.h
```diff
@@ -549,6 +549,54 @@ Primitive __device__ __host__ __inline__ Calc_Interface_Linear(Primitive const &
...
+  output.velocity_x = interface(cell_i.velocity_x, cell_im1.velocity_x, slopes_i.velocity_x, slopes_im1.velocity_x);
+  output.velocity_y = interface(cell_i.velocity_y, cell_im1.velocity_y, slopes_i.velocity_y, slopes_im1.velocity_y);
+  output.velocity_z = interface(cell_i.velocity_z, cell_im1.velocity_z, slopes_i.velocity_z, slopes_im1.velocity_z);
+  output.pressure   = interface(cell_i.pressure, cell_im1.pressure, slopes_i.pressure, slopes_im1.pressure);
+
+#ifdef MHD
+  output.magnetic_y = interface(cell_i.magnetic_y, cell_im1.magnetic_y, slopes_i.magnetic_y, slopes_im1.magnetic_y);
+  ;
+  output.magnetic_z = interface(cell_i.magnetic_z, cell_im1.magnetic_z, slopes_i.magnetic_z, slopes_im1.magnetic_z);
+  ;
```
ojwg: are these ';' needed?

src/reconstruction/reconstruction.h
```diff
@@ -241,100 +257,127 @@ Primitive __device__ __host__ __inline__ Van_Leer_Slope(Primitive const &left_sl
...
  */
-Characteristic __device__ __inline__ Primitive_To_Characteristic(Primitive const &primitive,
-                                                                 Primitive const &primitive_slope,
-                                                                 Real const &sound_speed,
-                                                                 Real const &sound_speed_squared, Real const &gamma)
-{
-  Characteristic output;
-
 #ifdef MHD
+eigenVecs __device__ __inline__ Compute_Eigenvectors(Primitive const &primitive, Real const &sound_speed,
```
ojwg, line 269: minor style issue -> capitalize eigenVecs

src/reconstruction/reconstruction.h
```diff
@@ -241,100 +257,127 @@ Primitive __device__ __host__ __inline__ Van_Leer_Slope(Primitive const &left_sl
...
+      numerator_2      = (output.magnetosonic_speed_fast_squared - sound_speed_squared);
       denom <= 0.0 or numerator_2 <= 0.0) {
-    alpha_fast = 1.0;
-    alpha_slow = 0.0;
-  } else if (Real const numerator_1 = (sound_speed_squared - magnetosonic_speed_slow_squared); numerator_1 <= 0.0) {
-    alpha_fast = 0.0;
-    alpha_slow = 1.0;
+    output.alpha_fast = 1.0;
+    output.alpha_slow = 0.0;
+  } else if (Real const numerator_1 = (sound_speed_squared - output.magnetosonic_speed_slow_squared);
```
ojwg, line 290: Since you're not using `numerator_1` in the scope of the if statement, I wonder if
`if ( (sound_speed_squared - output.magnetosonic_speed_slow_squared) <= 0.0) {` 
might be more readable?

ojwg (approved):
Looks good.

src/reconstruction/reconstruction.h
bcaddy: I think this is fixed in the latest commit

src/reconstruction/reconstruction.h
bcaddy: `numerator_1` is used in the `else` part of the statement. These declartions are a little weird in that they're defined for the entire if, else if, else block and not confined to a single part of it.

src/reconstruction/reconstruction.h
bcaddy: Fixed

PR #309, opened 2023-07-03
Enable clang-tidy checks

bcaddy:
## hicpp-signed-bitwise 

This check only failed in a few spots and it was with string literals operating with an unsigned int types so I set the
IgnorePositiveIntegerLiterals option to true.

## misc-confusable-identifiers check

The only thing that triggered this was `a1` and `al` in the Roe solver. It turns out the expression for `al` was short and only used one place so I copied and pasted it there, eliminating `al` entirely. I did the same for `ar` to maintain symmetry.

## readability-identifier-naming

Fixed and enabled name case checking for:

- namespaces
- enums
- macro definitions
- typdefs
- type aliases

## cppcoreguidelines-pro-bounds-constant-array-index

Disabled. This would be good to have on but it's just not feasible in GPU code since it pretty much requires heavy use of the STL.


bcaddy:
This should be ready to merge. I would like someone to take a look at the changes in commit [45ace1b](https://github.com/cholla-hydro/cholla/pull/309/commits/45ace1b794caa7799378f098fbc1bb6d1df5e136) since I changed the logic a bit. It should yield the same results, be clearer, and provide warning on a failure but I want a second set of eyes just in case.

PR #310, opened 2023-07-11
Turn on clang-tidy check for function naming and change function names to pass check

helenarichie:
For some reason that's not currently clear to me, clang-tidy only catches some instances of function names not conforming to the `Upper_Snake_Case` naming convention that we've chosen. I've corrected all of the functions that the check was failing for (and others that I happened to catch while doing this) and the check now passes. 

For the most part, it was straightforward to choose new versions of the function names to conform to our naming standard, but there are a few cases that I could use a second opinion on. In particular, there were a few function names in models/disk_ICs.cpp that I didn't understand the meaning of, so the new names that I gave them might not be the best.


bcaddy:
All the MHD and testing stuff looks good to me.

src/gravity/paris/ParisPeriodic.cu
```diff
@@ -16,9 +16,9 @@ ParisPeriodic::ParisPeriodic(const int n[3], const double lo[3], const double hi
       ddk_(2.0 * double(n[2] - 1) / (hi[2] - lo[2])),
   #elif defined PARIS_5PT
       nk_(n[2]),
-      ddi_(sqr(double(n[0] - 1) / (hi[0] - lo[0])) / 6.0),
-      ddj_(sqr(double(n[1] - 1) / (hi[1] - lo[1])) / 6.0),
-      ddk_(sqr(double(n[2] - 1) / (hi[2] - lo[2])) / 6.0),
+      ddi_(Sqr(double(n[0] - 1) / (hi[0] - lo[0])) / 6.0),
```
ojwg, line 19: `sqrt` is a library call, so this name change will break the code.

src/particles/feedback_CIC_gpu.cu
```diff
@@ -36,7 +36,7 @@ int snr_n;
 }  // namespace supernova
 
   #ifndef O_HIP
-__device__ double atomicMax(double* address, double val)
+__device__ double Atomic_Max(double* address, double val)
```
ojwg: HIP has `atomicMax`, so if compiling with O_HIP, we use it.  Changing this will break on AMD if you also update client code to call the new name.

ojwg (changes requested):
I think the disk_ICs function renaming is good, but there are a couple library function names that need to be changed back.

src/particles/feedback_CIC_gpu.cu
helenarichie: Resolved in [1465895](https://github.com/helenarichie/cholla/commit/146589509a5250e3280c3212889c52a31ed513c1)

src/gravity/paris/ParisPeriodic.cu
helenarichie: Maybe I'm misunderstanding something, but I thought those were calls to an inline function defined in [line 7](https://github.com/cholla-hydro/cholla/blob/5bdd42b8b728e4fdc39bfc783a3d38b4e88e239a/src/gravity/paris/ParisPeriodic.cu#L7) of this file that returns the square of the argument?

evaneschneider:
I think this one is good to go once it's merged with the updated version of dev.

helenarichie:
Okay, this is ready to go. I resolved the merge conflicts ([42fdb67](https://github.com/cholla-hydro/cholla/pull/310/commits/42fdb670c15b550188716d4a554484ac31acafaa)) and updated the function calls/new definitions to pass the function naming check ([2c0b6e2](https://github.com/cholla-hydro/cholla/pull/310/commits/2c0b6e2cb3d45af1557b6d2ffb4a45703e6d8ea3)). I also had to make a few extra commits for formatting and updating my test data submodule.

Issue #313, opened 2023-07-20
1D tests on dev don't run


evaneschneider:
When trying to run a test on the dev branch with ny = 1 and nz = 1 in a parameter file, the code exits with the following error:

Invalid boundary conditions. Must select between 1 (periodic), 2 (reflective), 3 (transmissive), 4 (custom), 5 (mpi).

This happens even though the boundary conditions are set in the parameter file, and appear to be read in, since the code correctly outputs:

Parameter values:  nx = 100, ny = 1, nz = 1, tout = 0.200000, init = Riemann, boundaries = 3 3 3 3 3 3

before it fails.

evaneschneider:
Removing the "if" statements in the Check_Custom_Boundary function in boundary_conditions.cpp fixes the boundary error, but the Riemann problems are still broken. The issue is that the initial conditions are not getting set, because the code never enters the loop.

bcaddy:
I fixed the underflow issue in the Riemann ICs (see PR #314). I'm not sure exactly what you want done in the boundary condition code so I didn't touch that. The tests still don't run though even with both fixes, after the first time step it returns all NaNs.

Once these issues are resolved we should add 1D and 2D sod tests to the system tests so we don't have this issue again.

alwinm:
Output from git bisect

Before this change, running 1-D sod.txt from examples completed successfully (though I don't guarantee numerical accuracy, I can say that the output from sod.txt has been consistent since main branch).

```
531efcbd2bbbd6b18f4b5a05f3e29f938f09a9ae is the first bad commit
commit 531efcbd2bbbd6b18f4b5a05f3e29f938f09a9ae
Author: Bob Caddy <r.Caddy@pitt.edu>
Date:   Tue Jun 20 15:27:40 2023 -0400

    Make all IC functions use parameter struct as arg
    
    Now all the initial condition functions just take the parameters struct
    as the argument instead of a bunch of Reals.

 src/grid/grid3D.h                     |  43 ++++++----
 src/grid/initial_conditions.cpp       | 153 +++++++++++++++++-----------------
 src/system_tests/mhd_system_tests.cpp |   3 +
 3 files changed, 104 insertions(+), 95 deletions(-)
 ```

alwinm:
With VL and bypassing the above bugs it runs. With SIMPLE and bypassing the above bugs it does not run. I think the differences between VL and SIMPLE warrant further investigation.

alwinm:
[...]

Issue #308, opened 2023-07-03
Fixing `clang-analyzer-core.UndefinedBinaryOperatorResult` with templates yields surprising performance improvement


bcaddy:
I was working on clang-tidy checks during the hack session today and ran into something interesting I think we should discuss. The `clang-analyzer-core.UndefinedBinaryOperatorResult` check checks for cases where a variable might be undefined when it's used and so structures like this tend to trigger it:

```cpp
  int o1, o2, o3;
  switch (dir) {
    case 0:
      o1 = grid_enum::momentum_x;
      o2 = grid_enum::momentum_y;
      o3 = grid_enum::momentum_z;
      break;
    case 1:
      o1 = grid_enum::momentum_y;
      o2 = grid_enum::momentum_z;
      o3 = grid_enum::momentum_x;
      break;
    case 2:
      o1 = grid_enum::momentum_z;
      o2 = grid_enum::momentum_x;
      o3 = grid_enum::momentum_y;
      break;
  }

// the check flags the usage of o1 on this line
velocity_x = dev_conserved[o1 * n_cells + id];
```

This is technically correct since `o1` could be uninitialized if `dir` is not 0-2 even though we know it always will be in that range. An easy way to fix this is if all the kernels with a structure like this take `dir` as a template argument instead of a regular argument, thus allowing the switch statement to be determined at compile time. 

While I don't think it's worth doing that just for this check I tried it out on a single kernel (the new PPMC_VL kernel) since templates can improve the performance of control statements by evaluating them at compile time. While I didn't expect it to help it actually did help. It reduced the execution time by ~4ms per time step, a ~2% improvement. If these results hold when applied to other kernels that could be an easy 10% performance improvement for very little effort. What do you all think?

bcaddy:
@evaneschneider, @alwinm, I'm especially interested in your takes on this

PR #315, opened 2023-07-22
Revert minor changes so 1-D examples can run and add debug functions

alwinm:
This PR contains a number of reverts as a hotfix to get 1-D examples (#313) working with minimal effort but will benefit from additional work and thought down the line. Most pressing would be bringing ppmc and ppmc tests back into sync--the thread guards in this PR allow non-zero values in ghost cells but the test expects zero values in ghost cells.

Old note:

Just a draft for now. Related to issue #313. 

After some feverish hacking around to find the bug, I think this revert might be promising but it also might not be. I'll take another look later.

There is probably a clean way to refactor this for compatibility with 1-D 2-D and 3-D without a full revert.


bcaddy:
Is the issue that the thread guard in `PPMC_CTU` is too restrictive? If so I'm curious about why it works at all in 3D since that's basically just the 1D case 3 times.

alwinm:
> Is the issue that the thread guard in `PPMC_CTU` is too restrictive? If so I'm curious about why it works at all in 3D since that's basically just the 1D case 3 times.

I think so... in the 3-D case your version of the thread guard correctly isolates the Real cells. In the old version, for simplicity between 1-D, 2-D, and 3-D, it does some work on ghost cells (which is extra work in the 2-D and 3-D cases). I'm guessing this extra work is negligible, simply adding a number of threads, and avoids some if-statements for each thread.

Basically, for 1-D, the "real cells" do go from 0 to nx in the perpendicular directions, not 3 to nx-3.

bcaddy:
Got it, that makes sense. I'm with you that I think there's a better way to compute the limits but I'm at a conference at the moment and don't have time to think about it, I'll consider it more when I get back. We might want to add the dimensionality as a template argument, per issue #308 making the arguments that are known at compile time into template arguments saves a surprising amount of time.

Also, the time saved in 3D isn't negligible. At 256^3 cells the ghost cells make up about 9% of the entire grid on the GPU (PPM with hydro) and the more aggressive thread guard saves about 6% of the runtime per kernel launch.

src/grid/boundary_conditions.cpp
```diff
@@ -151,12 +151,15 @@ int Grid3D::Check_Custom_Boundary(int *flags, struct parameters P)
   }
 
   for (int i = 0; i < 6; i++) {
+    /* Alwin: I am disabling this check because it is needlessly occurring every timestep.
```
bcaddy, line 154: is this disabled always? Is the check occurring somewhere else? 

- I think it should happen somewhere, probably in the error checking function.
- If we don't want it here then it should be deleted, not commented out

src/grid/initial_conditions.cpp
```diff
@@ -508,9 +508,9 @@ void Grid3D::Square_Wave(parameters const &P)
  *  \brief Initialize the grid with a Riemann problem. */
 void Grid3D::Riemann(parameters const &P)
 {
-  size_t const istart = H.n_ghost;
-  size_t const iend   = H.nx - H.n_ghost;
-  size_t jstart, kstart, jend, kend;
+  int const istart = H.n_ghost;
```
bcaddy, line 511: Why are all these `int`? If the issue is with `size_t` underflowing then we should figure out different logic since we decided to move to using `size_t` for indexing.

src/grid/initial_conditions.cpp
```diff
@@ -582,12 +583,12 @@ void Grid3D::Riemann(parameters const &P)
...
 #ifdef DE
             C.GasEnergy[id] = P.P_r / (gama - 1.0);
-#endif  // DE
-          }
-        }
-      }
-    }
-  }
+#endif       // DE
+          }  // if diaph
```
bcaddy, line 587: I don't think we should have labels like this. They're likely not to stay up to date and most IDEs/editors include color coding to help with which brackets belong with which.

src/reconstruction/ppmc_cuda.cu
```diff
@@ -27,11 +27,44 @@ __global__ void PPMC_CTU(Real *dev_conserved, Real *dev_bounds_L, Real *dev_boun
...
+      zs = 2;
+      ze = nz - 3;
+      break;
+  }
+  if (xid < xs || xid >= xe || yid < ys || yid >= ye || zid < zs || zid >= ze) {
+    return;
+  }
+
+  /*
   // Ensure that we are only operating on cells that will be used
```
bcaddy, line 62: This should be deleted instead of commented out.

src/reconstruction/ppmc_cuda_tests.cu
```diff
@@ -24,6 +24,8 @@
 
 TEST(tHYDROPpmcCTUReconstructor, CorrectInputExpectCorrectOutput)
 {
+  // Alwin: skip until this has been fixed
+  GTEST_SKIP();
```
bcaddy, line 28: I'm happy to help fix this test but I don't think that just marking tests as skippable when we change things is a good idea.

bcaddy (changes requested):

src/reconstruction/ppmc_cuda_tests.cu
alwinm: Marking a test as skippable when the test passed an incorrect function and fails the correct function is a good idea

src/grid/boundary_conditions.cpp
bcaddy: Fixed in [297225b](https://github.com/cholla-hydro/cholla/pull/315/commits/297225bd5c6d786d278ac652adf14a8c5901a5bd)

src/grid/initial_conditions.cpp
bcaddy: Removed in [df7df09](https://github.com/cholla-hydro/cholla/pull/315/commits/df7df09ab36ce0757abec64009eb05af275dbeef)

src/grid/initial_conditions.cpp
bcaddy: Fixed in [a2e1844](https://github.com/cholla-hydro/cholla/pull/315/commits/a2e18445686171c368dfa9113bf03476d5d90cbf)

src/reconstruction/ppmc_cuda_tests.cu
bcaddy: Fixed in [11804ba](https://github.com/cholla-hydro/cholla/pull/315/commits/11804ba2362387f67143ca0f1f135b1a69856a8d)

bcaddy (approved):

bcaddy:
Per the offline discussion. I fixed the issues I noted. I also fixed this issue in other reconstruction+integrator combinations. To avoid code duplication I wrote a function for the header guard and added a test for it.

evaneschneider:
Thank you for making those changes. I find the new thread guard function quite hard to read, could you add some more comments explaining to the user what is going on there? I.e. explain what "order" is, and what checks are actually happening?

bcaddy:
Done.

I also added truly 1D and 2D tests so that hopefully this doesn't happen again.

evaneschneider:
Good call!

PR #321, opened 2023-08-24
Misc. Small Changes for MHD timing

bcaddy:
- Allow MHD to build with the `OUTPUT` and `HDF5` flags absent
- Permit single line if statements without braces in clang-tidy
- Minor refactor of the `Time::Print_Average_Times` to make future changes to the header easier
- Add a note about the reduced range of the timers to `run_timing.log`


alwinm (approved):
looks good to me

PR #322, opened 2023-08-25
Print out the amount of GPU memory used at the end of each time step

bcaddy:
# Summary

When `CPU_TIME` is enabled Cholla now prints out how much GPU memory is allocated at the end of each time step. To do this I added new functions, `Reduce_size_t_Max` an `Print_GPU_Memory_Usage`, to determine the maximum amount of memory used on any GPU and print that information out. The motivation was to determine how many cells we can fit on a given GPU with a given Cholla configuration easily. Early tests show that on Frontier we can fit about 457^3 cells per GPU in an MHD only build.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -14,7 +14,9 @@
 #include "global/global.h"
 #include "grid/grid3D.h"
 #include "io/io.h"
+#include "utils/cuda_utilities.h"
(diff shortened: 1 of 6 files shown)


evaneschneider:
Looks good!

PR #327, opened 2023-09-07
Ensures that any directories specifed as part of ``outdir`` exist.

mabruzzo:
This is accomplished through the new ``Ensure_Outdir_Exists`` function.

Along the way, I introduced a helper function ``Is_Root_Proc()``, local to ``io.cpp`` so that we can reduce some of the ``ifdef`` statements. If you don't like this, I'm happy to remove it.

### DIFF
--- a/src/io/io.cpp
+++ b/src/io/io.cpp
@@ -10,7 +10,6 @@
 #include <fstream>
 #include <iostream>
 #include <string>
-#include <vector>
 #ifdef HDF5
   #include <hdf5.h>
 #endif  // HDF5
@@ -47,7 +46,7 @@ static inline bool Is_Root_Proc()
 
 void Create_Log_File(struct parameters P)
 {
-  if (!Is_Root_Proc()) {
+  if (not Is_Root_Proc()) {
     return;
   }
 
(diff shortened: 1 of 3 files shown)


src/io/io.cpp
```diff
@@ -6,9 +6,11 @@
 
 #include <algorithm>
 #include <ctime>
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #include <string>
+#include <vector>
```
alwinm, line 13: Basic question, what is <vector> providing in this PR?

src/io/io.cpp
```diff
@@ -64,11 +75,9 @@ void Create_Log_File(struct parameters P)
 
 void Write_Message_To_Log_File(const char *message)
 {
-#ifdef MPI_CHOLLA
-  if (procID != 0) {
+  if (!!Is_Root_Proc()) {
```
bcaddy, line 78: Is the double negation intended? 

Also, several months ago there was a discussion on using keyword vs. character Boolean operators (! vs not, || vs or, etc). If I recall correctly we decided to use keyword operators.

src/io/io.cpp
mabruzzo: that was an oversight. I thought I removed it. I'll fix it now

src/io/io.cpp
mabruzzo: The double negative was definitely an oversight. Thanks for catching that.

Gotcha, I'll shift to using `not` and `or`

PR #329, opened 2023-09-11
Allow KH_res_ind_2D initial conditions to be set

evazlimen:
KH_res_ind() was always setting the 3D initial conditions. I added an if/else statement so that 3D initial conditions are only set if there is more than one z grid cell. 

- KH_res_ind() sets 2D ICs for nz =1 and 3D ICs for nz >1
- Reduced standard runtime in the KH_res_ind_3D parameter file to t=3. Evolution is very non-linear by the old runtime of t=5 and it is not really resolution independent at that time. Up to t=3 it stays in the linear regime.
- Reduced the initial velocities for 2D and 3D KH_res_ind ICs to the velocities in KH_discontinuous_2D on advice of @brantr. Keeps this example problem simpler and makes it more resolution independent. 
I could not get clang to work on my computer sorry!

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -37,8 +37,7 @@ Checks: "*,
 
         google-readability-avoid-underscore-in-googletest-name,
         google-upgrade-googletest-case,
--*,
-readability-braces-around-statements,
+
         -bugprone-implicit-widening-of-multiplication-result,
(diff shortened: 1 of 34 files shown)


evazlimen:
I was able to fix the clang-format issues :)

PR #335, opened 2023-09-14
Slight upgrades to s99_feedback

alwinm:
Added some potentially useful debug functions (for feedback)
Added TI_cool
Refactored cooling for readability and easier reasoning (this change imo should also go into dev eventually)
Added functions for applying velocity and temperature ceilings


evaneschneider:
I'm curious - the clang-tidy checks that are failing for this are mostly about if statements not being inside braces (which I think we are okay with?), but there are a few that I don't really understand, primarily in the loading of the cloudy textures and a couple in disk galaxy and feedback. Do you @alwinm and @ojwg have a good understanding of the errors?

alwinm:
> I'm curious - the clang-tidy checks that are failing for this are mostly about if statements not being inside braces (which I think we are okay with?), but there are a few that I don't really understand, primarily in the loading of the cloudy textures and a couple in disk galaxy and feedback. Do you @alwinm and @ojwg have a good understanding of the errors?

I don't have a strong understanding of the errors but I believe they are already in dev. I am guessing that the reason why these lines are not giving us clang tidy errors in dev is because the Maketypes are slightly different and do not include them.

bcaddy:
Most of the more confusing errors I've already fixed in dev and the single line if statements without braces are also allowed in dev now (I think). You can either rebase or consult the changes to those sections in dev.

evaneschneider:
Yeah, unfortunately we can't rebase this from the current dev because something in there is breaking the simulations (that's why this PR is to a new branch). If the errors aren't causing problems I will merge this in.

PR #347, opened 2023-10-01
Refactor Python Concatenation Scripts + Add Dask Templates 

bcaddy:
## `cat_slice.py` Refactor

In a similar vein to my refactor of `cat_dset_3d.py` awhile ago I refactored `cat_slice.py` to be easier to use without editing and to allow people to import it into their scripts. I added a CLI and an internal function `concat_slice` that concatenates a single output time. This means that this function can be easily called in parallel with a tool like Dask when concatenating many slices. I also eliminated the intermediate arrays so that we wouldn't run out of memory when concatenating.

There are also additional options for compression, what datatype to save as, etc.

## `cat_dset_3d.py` Refactor

A minor refactor here. Moved the code for concatenating a single output time into its own function (`concat_3d_single`) so that it can be called in parallel with Dask like `concat_slice`. Also, added arguments for compression, what datatype to save as, and which fields to skip.

## Dask Templates

Added two templates for using Dask, one for running on a single machine and one for a distributed machine. The later also includes a Slurm script that works on Andes and Crusher; presumably Frontier as well though I haven't tested that. If you want an example of Dask in practice on Cholla data see the `dask-local-runner.py` and `dask-andes-runner.py` scripts in my [JSI Talk repo](https://github.com/bcaddy/JSI_Workshop_Talk_2023/tree/main/python).


python_scripts/cat_slice.py
helenarichie, line 1: Do you have an example of what it would look like to run this from the command line? I imagine it's something like `python cat_slice.py -s ./ -o ./ -n 4 -t 0`, but it would be useful to have that in the header of this file.

helenarichie: I'll note that I tried testing it out some data I have on ppc-n0 and ran into the following error:

```
[her45@ppc-n0 2023-10-20]$ python cat_slice.py -s ./hdf5/raw/ -o ./hdf5/slice/ -n 4 -t 0
Traceback (most recent call last):
  File "cat_slice.py", line 209, in <module>
    main()
  File "cat_slice.py", line 49, in main
    num_ranks=args.num_processses,
AttributeError: 'Namespace' object has no attribute 'num_processses'
```

Maybe I'm doing something wrong?

python_scripts/cat_dset_3D.py
helenarichie, line 1: Hm, it seems I'm also unable to get this one to run. Again, let me know if I'm doing something wrong.

```
[her45@ppc-n0 2023-10-20]$ python cat_dset_3D.py -s 0 -e 1 -n 4 -i ./hdf5/raw/ -o ./hdf5/full/
Traceback (most recent call last):
  File "/ihome/crc/install/power9/python/anaconda3-2019.10/lib/python3.7/site-packages/h5py/_hl/files.py", line 154, in make_fid
    fid = h5f.open(name, h5f.ACC_RDWR, fapl=fapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 78, in h5py.h5f.open
OSError: Unable to open file (unable to open file: name = 'hdf5/full/0.h5', errno = 2, error message = 'No such file or directory', flags = 1, o_flags = 2)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "cat_dset_3D.py", line 186, in <module>
    main()
  File "cat_dset_3D.py", line 49, in main
    compression_options=args.compression_opts)
  File "cat_dset_3D.py", line 92, in concat_3d
    compression_options=compression_options)
  File "cat_dset_3D.py", line 124, in concat_3d_single
    fileout = h5py.File(output_dir / f'{output_number}.h5', 'a')
  File "/ihome/crc/install/power9/python/anaconda3-2019.10/lib/python3.7/site-packages/h5py/_hl/files.py", line 312, in __init__
    fid = make_fid(name, mode, userblock_size, fapl, swmr=swmr)
  File "/ihome/crc/install/power9/python/anaconda3-2019.10/lib/python3.7/site-packages/h5py/_hl/files.py", line 156, in make_fid
    fid = h5f.create(name, h5f.ACC_EXCL, fapl=fapl, fcpl=fcpl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5f.pyx", line 98, in h5py.h5f.create
OSError: Unable to create file (unable to open file: name = 'hdf5/full/0.h5', errno = 2, error message = 'No such file or directory', flags = 15, o_flags = c2)
```

helenarichie:
I really like these scripts, I think they're going to be very useful. I just have a few comments and questions.

- I see that the arguments/general structure are slightly different between cat_dset_3d and cat_slice (e.g. having concat_3d and concat_3d_single for full-volume data vs just having concat single be the default behavior for cat_slice). Is it necessary that they aren't consistent? I think it'd be better if they were, but at least having the same names of command line arguments between the two would make these tools easier to use. I also think that having the option to specify a sequence of numbers using -s and -e arguments from the command like like you can in cat_dset_3d is functionality I'd miss in cat_slice.
- Just a minor suggestion, but I like cat_slice and cat_full and then maybe eventually cat_proj for the names of these instead of cat_dset_3d. I think it makes it clearer that they all do generally the same thing, but feel free to ignore if you don't agree.

Other than that, I think this is all great.

python_scripts/cat_dset_3D.py
helenarichie, line 1: I tested this one out and it works great!

evaneschneider:
Agreed this is a great refactor. Along the lines of Helena's comments, could you update the wiki documentation to be a little bit more clear about how to use these? For example, it is not clear from the example code in the wiki if the default behavior is to automatically loop through all the slices for a given snapshot in a directory (i.e. is the code somehow checking for the number of processes, or is that an argument?), or if you comment there referred to looping through all snapshots.

python_scripts/cat_slice.py
bcaddy: You weren't doing anything wrong, I misspelled the variable name. It's fixed and will be pushed shortly.

bcaddy:
@helenarichie, I'll work on unifying the API and structure between the two. Good catch, I did them at separate times and didn't realize how different they are.

@evaneschneider and @helenarichie, I agree that the naming of the files is a bit confusing and is something I found confusing when I first saw them. We currently have 6 unique scripts for concatenating data files in varying states without consistency. I think I'll open an issue for unifying them all under a more common structure and API but I have a few questions:

- `cat_dset_3d.py` IMO is unclear compared to the rest. How does `cat_3d_dataset.py` sound? That's more consistent with the other ones
- `cat.py` doesn't appear to be used, can I delete it? I think that a shared library is good, we're just not using it as is.

bcaddy:
@helenarichie, I think I've addressed your comments except for the naming. Could you look over it again?

bcaddy:
@evaneschneider, in the wiki. Are you talking about the section at the end of the "Outputs" page? I didn't write that, I think Alwin did. His code in `cat.py` is importable functions like what I've written here but without the CLI that these scripts have. He and I have talked offline and decided that my new scripts are probably a better option so I think next hack day I'll update the other concatenation scripts to be similar to mine, update the wiki, and remove `cat.py`. Does that sound good?

alwinm:
> Agreed this is a great refactor. Along the lines of Helena's comments, could you update the wiki documentation to be a little bit more clear about how to use these? For example, it is not clear from the example code in the wiki if the default behavior is to automatically loop through all the slices for a given snapshot in a directory (i.e. is the code somehow checking for the number of processes, or is that an argument?), or if you comment there referred to looping through all snapshots.

I wrote cat.py and that section of the Wiki so that each solo function would detect the number of processes without argument: the code detects number of processes. The while loop in the 2nd section goes through all snapshots aka output indices *.h5 assuming the output interval for that data type is 1.

evaneschneider:
> @evaneschneider, in the wiki. Are you talking about the section at the end of the "Outputs" page? I didn't write that, I think Alwin did. His code in `cat.py` is importable functions like what I've written here but without the CLI that these scripts have. When I started on my version I didn't realize another importable version existed. He and I have talked offline and decided that my new scripts are probably a better option in the long run so I think next hack day I'll update the other concatenation scripts to be similar to mine, update the wiki, and remove `cat.py`. Does that sound good?

That sounds fine to me. I'll merge this in once the documentation has been updated.

helenarichie (approved):

helenarichie:
These changes all look great to me!

bcaddy:
This is ready to review. There have been some significant updates since the last review so, if you have time, I'd like @helenarichie to review it again. Brief summary of changes below

- Add support for projections and rotated projections to slice concatenation script
- Rename scripts to `concat_2d_data.py` and `concat_3d_data.py` for clarity and consistency
- Move a lot of common functions into separate file (`/concat_internals.py`)
- Update the concatenation script for particles with new method
- Remove deprecated `cat.py` file
- Update documentation in wiki (@evaneschneider, this is live, could you take a look?)

python_scripts/concat_particles.py
```diff
@@ -0,0 +1,250 @@
...
+def __setup_destination_file(source_directory: pathlib.Path,
+                             destination_file: h5py.File,
+                             output_number: int,
+                             num_particles: int,
+                             skip_fields: list,
+                             destination_dtype: np.dtype,
+                             compression_type: str,
+                             compression_options: str,
+                             chunking) -> h5py.File:
+  """_summary_
```
helenarichie, line 174: Is this meant to have a short description of what this function does?

helenarichie (approved):
This all looks fantastic! I noticed that part of a doc-string might be incomplete, but, otherwise, this is ready to go. I took a look at the documentation as well and I think it's clear and has everything it needs.

python_scripts/concat_particles.py
bcaddy: Good catch. Fixed it. Thank you.

Issue #286, opened 2023-04-10
Multiple versions of cuda error checking with unclear performance impacts


bcaddy:
There's 3 versions of cuda error checking in `global_cuda.h`, `gpuErrchk`, `CudaSafeCall`, and `CudaCheckError`. They all do pretty much the same thing but with small differences, some unclear performance impacts, different syntax/usage, and are or are not behind the `CUDA_ERROR_CHECK` ifdef. I think we should merge these into one, clarify how they should be used, and put only the expensive `cudaDeviceSynchronize` behind an ifdef. 

I would like some input on the best way to do this and what potential traps await.

alwinm:
Some of my preliminary thoughts on the matter:

I like the version where it accurately tells you which file and line invoked the problem using the __line__ and __file__ macros.

Synchronize is necessary to make it actually useful for kernels, but not necessary for things like cudamemcpy which have their own synchronize... but agreed that it should behind an ifdef. In the grand scheme of things it is not prohibitively expensive but I do agree that it's usually an unnecessary cost. Having a single version will help ensure that all such calls uniformly toggle synchronize. 

In terms of naming we should probably move towards gpu rather than cuda, and CheckError rather than SafeCall (since it is not really guaranteeing safety...? unless I am misunderstanding something).

bcaddy:
Currently it looks like `gpuErrchk` isn't actually used anywhere and without `CUDA_ERROR_CHECK` the other two do absolutely nothing which is problematic. Here's my psuedo-code proposal that I think works for all use cases and will actually check things 

```cpp
#define gpuCheckError(code) { gpuAssert((code), __FILE__, __LINE__); }
inline void gpuAssert(cudaError_t code = cudaPeekAtLastError(), const char *file, int line, bool abort=true)
{
  #ifdef CUDA_ERROR_CHECK
  code = cudaDeviceSynchronize();
  #endif

   if (code != cudaSuccess) 
   {
      fprintf(stderr,"GPUassert: failed at %s:%i with code %s\n", file, line, cudaGetErrorString(code));
      if (abort) exit(code);
   }
}
```

I'm not sure it the argument defaulting will work with the macro, if not we might need two macros, one for checking kernel launches and one for everything else.

alwinm:
I know you said its pseudocode, but I recommend replacing exit with chexit, and we should think about what we really want to do with the print statement if multiple MPI ranks fail.

bcaddy:
I edited the code to add `chexit`.

[...]

Issue #296, opened 2023-05-10
What should we do about device memory checking?


alwinm:
With -DCUDA_ERROR_CHECK, all cudaMallocs which run out of memory return an error code as expected on both Nvidia and AMD (ppc-n0 and Crusher), and execution is halted. 

Without -DCUDA_ERROR_CHECK, cudaMallocs which run out of memory silently fail. On Nvidia, it may continue to run but do nothing on each timestep. On AMD, memory access fault. 

In either case, the returned pointer is NULL, which is an easy way to check for failure.

I think there are multiple valid options:

1. Leave as is. Users should be expected to understand their memory requirements and/or turn on CUDA_ERROR_CHECK when debugging. 
2. Add a util which prints gpu memory usage to help with the above. 
3. Replace cudaMalloc with a macro gpuMalloc which calls cudaMalloc and always does a check, whether or not CUDA_ERROR_CHECK is on, printing LINE/FILE and calling ```chexit(-1)``` upon failure. The check could be one or multiple of the following:
```
pointer != NULL
```
or
```
cudaError err == cudaGetLastError();
err == cudaSuccess
```
or 
```
cudaError err = cudaMalloc();
err == cudaSuccess
```

bcaddy:
I generally like option 2 more. I don't see how option 3 works without setting `err` to something.

I would argue that this is an extension of issue #286 and we should address both at once. I added a proposal to that PR on what code we should use

bcaddy:
Also, note that `cudaGetLastError()` and `cudaPeekAtLastError()` aren't the same. The `Get` version resets the error to `cudaSuccess` after it's done and the `Peek` version doesn't. My guess is that we want the former since we don't actually do any handling of errors and instead just exit. [source](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__ERROR.html#group__CUDART__ERROR_1g3529f94cb530a83a76613616782bd233)

alwinm:
> I generally like option 2 more. I don't see how option 3 works without setting `err` to something.
> 
> I would argue that this is an extension of issue #286 and we should address both at once. I added a proposal to that PR on what code we should use

I do not consider this an extension of issue #286 because cudaMalloc is special:

- It has its own sync, does not require cudaDeviceSynchronize
- It returns its own error code, does not require cudaGetLastError. 
- We could consider treating running out of memory as a special error that we always want to check, whether or not CUDA_ERROR_CHECK is on.

[...]

PR #350, opened 2023-10-27
Consolidate GPU Error Checking Function

bcaddy:
## GPU Error Checking

I replaced the 3 macros and 7 functions for GPU error checking with a single overloaded function; one overload for CUDA/HIP checking and one for CUFFT/HIPFFT checking. The function supports wrapping a CUDA call or being called with no arguments to check the latest error.

I also added error checking to some `cudaMallocs` that were missing them or used them in a non-standard way.

The other major change is the deprecation of the `CUDA_ERROR_CHECK ` macro. Now error checking is on by default and can be disabled with the new `DISABLE_GPU_ERROR_CHECKING` macro.

This should resolve Issue #286 and possibly #296 as well, subject to discussion in that issue.


mabruzzo (approved):
So I've gone through -- it looks good to me.

I assume that you have confirmed that any of the implicit grid-synchronizations performed by the old error checking macros were entirely unnecessary, right? If so, we're good to merge.

bcaddy:
Yep, everywhere that the implicit sync seemed like it might be relevant were on functions that already contain an implicit sync (like moving or allocating). Since we only use 1 GPU stream there's an implicit sync between all kernels.

PR #353, opened 2023-11-01
Introduce 2 tweaks to ``DeviceVector`` to improve safety of the class

mabruzzo:
1. Delete implicitly declared constructors/assignment operations. We could definitely define these, but currently they can lead to dereferencing null pointers or double-freeing pointers

2. While I was here, I also added an explicit check that the elements of the vector are trivially copyable. This is the formal requirement for being able to copy an object with a variant of memcpy.

### DIFF
--- a/src/utils/DeviceVector.h
+++ b/src/utils/DeviceVector.h
@@ -13,6 +13,7 @@
 #include <algorithm>
 #include <stdexcept>
 #include <string>
+#include <type_traits>
 #include <vector>
 
 // External Includes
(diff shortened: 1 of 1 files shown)


src/utils/DeviceVector.h
```diff
@@ -60,6 +65,15 @@ class DeviceVector
    */
   ~DeviceVector() { _deAllocate(); }
 
+  /* The following are deleted because they currently lead to invalid state.
+   * (But they can all easily be implemented in the future).
+   */
+  DeviceVector()                                           = delete;
```
bcaddy, line 71: Good catch!

bcaddy (approved):
Looks great to me, always happy to improve the safety of this class given that the whole point of it is a safer and easier device array. Other than the comment on the error message I think this is good to go.

PR #355, opened 2023-11-21
Convert `OUTPUT_ALWAYS` from compile-time definition to a run-time parameter called `output_always`

mabruzzo:
This is fairly self-explanatory (and I think the benefits are also self-explanatory).


src/global/global.h
```diff
@@ -179,6 +179,7 @@ struct Parameters {
 #ifdef DE
   int out_float32_GasEnergy = 0;
 #endif
+  int output_always = 0;
```
bcaddy, line 182: I might explicitly make this a bool rather than an int. Other than that looks great.

evaneschneider:
Currently failing clang-tidy, which wants 'false' if you go with a bool.

PR #360, opened 2023-12-15
Remove Deprecated CUDA Macro

bcaddy:
Closes issue #241


ojwg (approved):
Removes CUDA macro with many format changes that look correct.

PR #361, opened 2024-01-09
Reconstruction bug fixes

bcaddy:
## PLMC 

The monotonization after the calculation of the interfaces was missing so I added it.

## Utility Function

A couple of utility functions had `*` when they should be `+` in the calculation of vector magnitude squared. Added a function for computing that so the mistake won't happen again. This should fix some of the other weird issues that we've been running into.


evaneschneider:
This looks good to me.

Issue #362, opened 2024-01-10
Move location of temperature floor


evaneschneider:
Right now the temperature floor is applied as a kernel call at the very end of each hydro integrator. I propose that instead, it be added as a single function call in the Grid3D::Update_Hydro_Grid function. If added right after the current Execute_Hydro_Integrator function, it shouldn't change the existing logic at all, but would reduce the number of function calls significantly (and allow it to be used for 1D and 2D simulations, a functionality that is currently missing.) This would require adding a new wrapper (host) function to call the kernel (as was done for other functions that were moved out of the integrators, e.g. cooling or the time step calculation).

PR #363, opened 2024-01-13
Add scalar floor and refactor temperature and density floors

helenarichie:
In summary, this PR:
- Adds the ability to apply a floor to any scalar in the conserved array in `Apply_Scalar_Floor`.
    -  The particular scalar the floor is applied to is specified using the `field_num` argument of `Apply_Scalar_Floor`, and should be specified with its `grid_enum` index.
- Refactors the density, temperature, and scalar floor values to be set from the input parameter file.
    - Deletes the `DENS_FLOOR` and `TEMP_FLOOR` macros (the floors are instead stored in the Grid3D header as `density_floor`, `temperature_floor`, and `scalar_floor`).
- Adds a unit test for `Apply_Scalar_Floor`.
- Refactors where the scalar and temperature floor functions are called from based on Issue #362.


src/grid/grid3D.cpp
```diff
@@ -258,16 +258,22 @@ void Grid3D::Initialize(struct Parameters *P)
...
   H.density_floor = 0.0;
 #endif
 
-#ifdef TEMPERATURE_FLOOR
-  H.temperature_floor = TEMP_FLOOR;
+#ifdef SCALAR_FLOOR
+  H.scalar_floor = P->scalar_floor;
 #else
-  H.temperature_floor = 0.0;
+  H.scalar_floor = 0.0;
```
evaneschneider, line 276: While I understand this logic for the density and temperature floors, I think it doesn't actually apply for the scalar floor, since scalars can have negative values (think for example about the transverse velocities in the Riemann problem). In practice, it shouldn't matter, since you don't call the floor function unless you are applying a floor, in which case the value should have been specified in the parameter file. But I think it might make more sense rather than setting a default value of 0 here to just have the code throw a warning if the floor macro is turned on but no input value is supplied.

evaneschneider (changes requested):
New parameter defaults should be set in global.h, and the code should give a warning if any floor is turned on but the input parameter isn't set.

helenarichie:
> New parameter defaults should be set in global.h, and the code should give a warning if any floor is turned on but the input parameter isn't set.

Addressed in [3e1c195](https://github.com/cholla-hydro/cholla/pull/363/commits/3e1c1959f277272cb17c362ca1c4169d36ef8567)

evaneschneider (approved):

PR #365, opened 2024-01-22
Add a check to indicate that MHD does not support projections

bcaddy:


evaneschneider:
I don't really understand this. Why can't projections be output when MHD is turned on?

bcaddy:
I haven't added support for it to the magnetic fields so if you run an MHD sim with it you'll just get the hydro fields which you probably don't expect. I'm happy to convert this to a warning instead of an error.

I can put adding support for projections to my to-do list if you want.

bcaddy:
I'll switch it to a warning. I'd like to have some kind of notification baked in about it so folks don't expect something they don't get and then get confused about it.

Slices support MHD using cell centered magnetic fields.

bcaddy:
wait, don't the temperature calculations need some tweaks to support MHD?

evaneschneider:
Yes, good point. So probably the clearest thing is to specify in the warning that the temperature projections don't take into account the magnetic pressure. (I still don't think this is that big a deal, since they're density weighted and unlikely to be used as anything other than a qualitative output.)

bcaddy:
Would it be that hard to add the magnetic pressure to the temperature calculations? I assume it's just a couple of lines

evaneschneider:
I think it would be extremely easy. I assumed this PR meant you didn't feel like it!

bcaddy:
I did not dig that deep, I just noticed when writing the MHD wiki page that projections weren't supported and we should probably error/warn on that. I'll look into it tomorrow

PR #366, opened 2024-01-23
Add MHD Support to Projection and Rotated Projection Outputs

bcaddy:
## Summary

Per the discussion in PR #365 I've added MHD support to the projection and rotated projection outputs along with a minor refactor to those functions to accomplish this and utilize more of the utility functions we've written.

I also added a utility function for computing the temperature from the conserved variables (`hydro_utilities::Calc_Temp_Conserved`). Internally it simply calls `hydro_utilities::Calc_Pressure_Conserved` and `hydro_utilities::Calc_Temp` to compute the temperature.

## Duel Energy + Dust Fixes

`Dust_Kernel` had few DE related bugs in it that I fixed with help from @helenarichie. While doing that I also slightly redefined `hydro_utilities::Calc_Temp_DE` to take the total gas energy instead of the specific gas energy and density.

### DIFF
--- a/src/dust/dust_cuda.cu
+++ b/src/dust/dust_cuda.cu
@@ -72,8 +72,8 @@ __global__ void Dust_Kernel(Real *dev_conserved, int nx, int ny, int nz, int n_g
 
     // Compute the temperature
   #ifdef DE
-    Real const total_gas_energy = dev_conserved[id + n_cells * grid_enum::GasEnergy];
-    Real const temperature      = hydro_utilities::Calc_Temp_DE(total_gas_energy, gamma, number_density);
+    Real const gas_energy  = dev_conserved[id + n_cells * grid_enum::GasEnergy];
+    Real const temperature = hydro_utilities::Calc_Temp_DE(gas_energy, gamma, number_density);
   #else  // DE is not enabled
     Real const energy     = dev_conserved[id + n_cells * grid_enum::Energy];
     Real const momentum_x = dev_conserved[id + n_cells * grid_enum::momentum_x];
--- a/src/dust/dust_cuda.cu
+++ b/src/dust/dust_cuda.cu
@@ -83,14 +83,15 @@ __global__ void Dust_Kernel(Real *dev_conserved, int nx, int ny, int nz, int n_g
     #ifdef MHD
     auto const [magnetic_x, magnetic_y, magnetic_z] =
         mhd::utils::cellCenteredMagneticFields(C.host, id, xid, yid, zid, H.n_cells, H.nx, H.ny);
-    #else   // MHD is not defined
-    Real const magnetic_x = 0.0, magnetic_y = 0.0, magnetic_z = 0.0;
-    #endif  // MHD
-
     Real const temperature =
         hydro_utilities::Calc_Temp_Conserved(energy, density_gas, momentum_x, momentum_y, momentum_z, gamma,
                                              number_density, magnetic_x, magnetic_y, magnetic_z);
-  #endif    // DE
+    #else   // MHD is not defined
+    Real const temperature = hydro_utilities::Calc_Temp_Conserved(energy, density_gas, momentum_x, momentum_y,
+                                                                  momentum_z, gamma, number_density);
+    #endif  // MHD
+
+  #endif  // DE
 
     Real tau_sp = Calc_Sputtering_Timescale(number_density, temperature, grain_radius) /
                   TIME_UNIT;  // sputtering timescale, kyr (sim units)
--- a/src/dust/dust_cuda.cu
+++ b/src/dust/dust_cuda.cu
@@ -52,14 +52,9 @@ __global__ void Dust_Kernel(Real *dev_conserved, int nx, int ny, int nz, int n_g
   int id_x    = id - id_z * nx * ny - id_y * nx;
 
   // define physics variables
-  Real density_gas, density_dust;           // fluid mass densities
-  Real number_density;                      // gas number density
-  Real mu = 0.6;                            // mean molecular weight
-  Real temperature, energy, pressure;       // temperature, energy, pressure
-  Real velocity_x, velocity_y, velocity_z;  // velocities
-  #ifdef DE
-  Real energy_gas;
-  #endif  // DE
+  Real density_gas, density_dust;  // fluid mass densities
+  Real number_density;             // gas number density
+  Real mu = 0.6;                   // mean molecular weight
 
   // define integration variables
   Real dd_dt;          // instantaneous rate of change in dust density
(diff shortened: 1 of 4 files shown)


src/io/io.cpp
```diff
@@ -1563,37 +1563,51 @@ void Grid3D::Write_Projection_HDF5(hid_t file_id)
...
+              mhd::utils::cellCenteredMagneticFields(C.host, id, xid, yid, zid, H.n_cells, H.nx, H.ny);
+    #else   // MHD is not defined
+          Real const magnetic_x = 0.0, magnetic_y = 0.0, magnetic_z = 0.0;
+    #endif  // MHD
+
+          Real const T = hydro_utilities::Calc_Temp_Conserved(E, C.density[id], mx, my, mz, gama, n, magnetic_x,
+                                                              magnetic_y, magnetic_z);
+  #endif    // DE
+
+          Txy += T * H.dz;
```
evaneschneider, line 1608: This line is missing the density-weighting (Cholla outputs density-weighted temperature projections)

src/io/io.cpp
```diff
@@ -1603,37 +1617,46 @@ void Grid3D::Write_Projection_HDF5(hid_t file_id)
...
-          Real my = C.momentum_y[id];
-          Real mz = C.momentum_z[id];
-          Real E  = C.Energy[id];
-          T       = (E - 0.5 * (mx * mx + my * my + mz * mz) / C.density[id]) * (gama - 1.0) * PRESSURE_UNIT / (n * KB);
-  #endif
   #ifdef DE
-          T = C.GasEnergy[id] * PRESSURE_UNIT * (gama - 1.0) / (n * KB);
-  #endif
-          Txz += T * C.density[id] * H.dy;
+          Real const T = hydro_utilities::Calc_Temp_DE(C.density[id], C.GasEnergy[id], gama, n);
```
evaneschneider, line 1640: Has n been calculated? Looks like maybe the calculation was accidentally deleted in this loop.

src/io/io.cpp
```diff
@@ -1603,37 +1617,46 @@ void Grid3D::Write_Projection_HDF5(hid_t file_id)
...
+          auto const [magnetic_x, magnetic_y, magnetic_z] =
+              mhd::utils::cellCenteredMagneticFields(C.host, id, xid, yid, zid, H.n_cells, H.nx, H.ny);
+    #else   // MHD is not defined
+          Real const magnetic_x = 0.0, magnetic_y = 0.0, magnetic_z = 0.0;
+    #endif  // MHD
+
+          Real const T = hydro_utilities::Calc_Temp_Conserved(E, C.density[id], mx, my, mz, gama, n, magnetic_x,
+                                                              magnetic_y, magnetic_z);
+  #endif    // DE
+          Txz += T * H.dy;
```
evaneschneider, line 1658: Missing density here too (unless I'm misunderstanding something)

src/io/io.cpp
bcaddy: Just added it back.

src/io/io.cpp
bcaddy: In that case I have a question. `hydro_utilities::Calc_Temp_DE` says that the DE temperature is 

`Real T = d * ge * (gamma - 1.0) * PRESSURE_UNIT / (n * KB);`

and the projection code has 

`Real T = C.GasEnergy[id] * PRESSURE_UNIT * (gama - 1.0) / (n * KB);`

Which differ by a factor of `d`. Is that a bug in `Calc_Temp_DE` or the projections?

evaneschneider: ge is usually defined as GasEnergy / density

bcaddy: Ok, my read is that the `Calc_Temp_DE` version is correct and the projection version should have the additional factor of `d` in it?

evaneschneider: Both are the same and are correct (both versions give you T). So yes, the projection version should be multiplied by density when you do the sum (not in the line where you calculate T). If you look at the previous version of the file, it's there, I just noticed you deleted it.

evaneschneider: The quantity GasEnergy in the conserved array is d*ge. Does that make sense?

bcaddy: Per an offline discussion with @evaneschneider, the confusion is in nomenclature. `GasEnergy` is the total thermal energy in a cell, `ge` is the specific thermal energy in a cell and they relate via the equation `GasEnergy = ge*density`.

src/io/io.cpp
bcaddy: fixed

src/dust/dust_cuda.cu
helenarichie, line 1: This looks great to me. Thanks again for these catches!

evaneschneider:
Is it your intention to keep a warning for rotated projections?

src/utils/hydro_utilities.h
```diff
@@ -63,11 +63,42 @@ inline __host__ __device__ Real Calc_Temp(Real const &P, Real const &n)
...
+  Real const P = Calc_Pressure_Conserved(E, d, mx, my, mz, gamma, magnetic_x, magnetic_y, magnetic_z);
+  return Calc_Temp(P, n);
+}
+
 #ifdef DE
-inline __host__ __device__ Real Calc_Temp_DE(Real const &d, Real const &ge, Real const &gamma, Real const &n)
+/*!
+ * \brief Compute the temperature when DE is turned on
+ *
+ * \param total_gas_energy The total gas energy in the cell. This is the value stored in the grid
```
helenarichie, line 94: Might be clearer to specifically say the name of the grid_enum index (as in grid_enum::Energy)

bcaddy: done

helenarichie: Also, I'm not sure that total_gas_energy is the clearest variable name. [In the style guide](https://github.com/cholla-hydro/cholla/discussions/206#discussioncomment-4429288), I think we decided on energy_tot to refer to the total gas energy (unless I'm misunderstanding something)

src/utils/hydro_utilities.h
```diff
@@ -63,11 +63,42 @@ inline __host__ __device__ Real Calc_Temp(Real const &P, Real const &n)
...
+ * \param gamma The adiabatic index
+ * \param n The number density
+ * \return Real The temperature
+ */
+inline __host__ __device__ Real Calc_Temp_DE(Real const &total_gas_energy, Real const &gamma, Real const &n)
 {
-  Real T = d * ge * (gamma - 1.0) * PRESSURE_UNIT / (n * KB);
-  return T;
+  return total_gas_energy * (gamma - 1.0) * PRESSURE_UNIT / (n * KB);
```
helenarichie, line 102: Same comment on the variable name here

bcaddy: The `total_gas_energy` variable? The previous comment is talking about that variable.

src/utils/hydro_utilities.h
evaneschneider: I don't think we want energy_tot, since that would imply the total energy, i.e. thermal + kinetic + magnetic etc. etc. My reading of the style guide is that it should be total_gas_energy.

evaneschneider: My impression is that the total is not actually necessary, since we usually mean the total when we say any of these (energy, kinetic energy, etc.) and really what we should be doing is clarifying when we are using the specific version (i.e. the old ge should be specific_gas_energy).

bcaddy: > I don't think we want energy_tot, since that would imply the total energy, i.e. thermal + kinetic + magnetic etc. etc. My reading of the style guide is that it should be total_gas_energy.
It is `total_gas_energy` I'm confused

> My impression is that the total is not actually necessary...
Sure, but I'd rather be extra clear since the ambiguity is what led to this multi-hour debacle.

evaneschneider: In fact, I don't see energy_total anywhere in the code, and would advocate for this approach. I think gas_energy is consistent with the way this is labeled elsewhere (i.e. in reconstruction.h).

helenarichie: Sorry, I think I created a bit of confusion because I was still editing my comments before I realized that you guys were already replying. However, I agree that the `total_` isn't that clear. I think the clearest thing to do is clarify in the doc string and follow what's done elsewhere in the code by calling it `gas_energy`.

bcaddy: I've changed the variable name to `gas_energy`.

evaneschneider (commented):
I think it would be more consistent if n and T were treated the same way between the two versions (projections versus rotated projections), since in the first function d and n and T are all redefined as Real const within the loop and in the second only T is, which makes them superficially appear to be doing different things. I also think it would be clearer if the non-MHD case didn't define the empty B-fields (as we discussed offline), so that it is clear the Calc_Temp function can be called without them. Otherwise, I think this is ready to go.

bcaddy:
I believe those things are done now.

bcaddy:
Everything is passing except a single clang-tidy warning in unrelated code. I don't know why that warning didn't pop up in a previous PR but it is fixed in PR #364. IMO, either that PR should be merged in first or this one is OK to merge.

PR #368, opened 2024-02-01
Make `AutomaticLaunchParams` Member Variables Private

bcaddy:
@ojwg  pointed out that `threadsPerBlock` and `numBlocks` were public, non-const, member variables of `AutomaticLaunchParams` which leaves them open to misuse. Now they're both private and are effectively read only through getters.


ojwg (approved):
Looks great

PR #369, opened 2024-02-01
Restart Tests in Jenkins Pipeline if they Fail

bcaddy:
~Work in progress. I just need this up so that I can actually test it~

It's ready now. If any of the tests fail they should retry once and I've already witnessed that working. The Jenkins docs are a little unclear about the meaning of the argument to the `retry` command so it might be set to retry twice instead of once, if we see that behavior we can just lower that argument to 1.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,12 +73,15 @@ pipeline
                     {
                         steps
                         {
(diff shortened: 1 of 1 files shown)


evaneschneider:
Works for me!

PR #370, opened 2024-02-02
Merge dev into main

evaneschneider:
This PR will merge the current version of dev into main, prior to the Cholla 3.0.0 release.


brantr:
Currently in `dev` there are some mismatches between `python_script/` example script filenames and what's described in the wiki. The `python_script/' examples also don't work with the new output directory organization. While these are minor issues, they may prevent users from actually successfully using the data. I think these could be addressed with a quick pr to `dev`, or they could be changed with a pr to `main` after you merge `dev`.

evaneschneider:
Thanks Brant! It was our intention to make everything consistent with the new directory structure - can you flag which ones we missed? Also, which wiki pages?

brantr:
Well, the scripts are `concat_3d_data.py` and the 2D version -- earlier today I swear they were not updated, but I see it now somehow says they were updated yesterday.  And I must have clicked on an old wiki or not refreshed because the revisions I wanted it says Helena made two weeks ago (again I checked this morning after notice of the dev->main PR).  So apologies for the false alarm.

evaneschneider:
No worries, I appreciate you taking a look! We've definitely been doing a lot of tidying the last couple of weeks. I'm going to go ahead and merge this PR, but feel free to submit issues about anything else you think should be addressed.

PR #371, opened 2024-02-06
Reconstruction Kernel Fusion 1: Convert to structs

bcaddy:
# Summary

The goal of this PR is a first step in preparing to fuse the reconstruction and Riemann solver kernels. To do this it converts many variables in both the reconstructions and the Riemann solvers into structs. Most of these structs were already present in the `reconstruction` and `mhd` namespaces but they have been generalized and moved into the `hydro_utilities` namespace in the `basic_structs.h` file and the `reconstruction` namespace in `reconstruction.h`.

Given the number of changes and the new structure of data I would like everyone to at least glance through the contents of `basic_structs.h` and `reconstruction.h` to see if it looks good to them since I plan on using those structs throughout the reconstructors.

### `hydro_utilitiies::Primitive`

Contains all the primitive variables and the specific versions of the gas energy and scalars. Included a default and non-default constructor, mostly the non-default constructor is for tests since the previous usage of this struct in tests had ABI dependencies; the new constructor provides a more flexible API to shield the internal arrangement of the struct.

### `hydro_utilitiies::Conserved`

Contains all the conserved variables and the non-specific versions of the gas energy and scalars. Also has a constructor for the reasons listed above.

### `hydro_utilitiies::Vector`

This just has three members, `x`, `y`, and `z`, and is intended to serve as an easy way to move vector quantities around. I don't love the name since it sort of conflicts with `std::vector` but they're in different namespaces and this one stores actual mathematical vectors. This struct is used in both the `Primitive` and `Conserved` structs to store their vector quantities. I would like some feedback on naming, usefulness, etc of this struct.

### `reconstruction::InterfaceState`

Contains all the variables used in the Riemann solver interface states. It's basically all the primitive and conserved variables together in one struct.

## Renamed `reconstruction.h` to `reconstruction_internals.h`

In the final version I want a file called `reconstruction.h` that only has the Riemann solver facing code and a separate one that has all the code that is shared between the reconstructors. This is why I renamed `reconstruction.h` and created a new `reconstruction.h` that contains only the `InterfaceState` struct at the moment.


src/utils/mhd_utilities.h
```diff
@@ -198,13 +199,15 @@ inline __host__ __device__ Real alfvenSpeed(Real const &magneticX, Real const &d
...
- * \return Real local struct with the X, Y, and Z cell centered magnetic
- * fields. Intended to be called with structured binding like `auto [x, y,
+ * \return hydro_utilities::Vector with the X, Y, and Z cell centered magnetic
+ * fields. Can be called with structured binding like `auto [x, y,
  * z] = mhd::utils::cellCenteredMagneticFields(*args*)
  */
-inline __host__ __device__ auto cellCenteredMagneticFields(Real const *dev_conserved, size_t const &id,
-                                                           size_t const &xid, size_t const &yid, size_t const &zid,
-                                                           size_t const &n_cells, size_t const &nx, size_t const &ny)
+inline __host__ __device__ hydro_utilities::Vector cellCenteredMagneticFields(Real const *dev_conserved,
```
helenarichie: This comment can be ignored if you want since it applies to a lot of the pre-existing code in this namespace, but the naming of this function and its returns don't follow the style guide.

src/utils/basic_structs.h
```diff
@@ -0,0 +1,94 @@
...
+#endif  // DE
+
+#ifdef SCALAR
+  Real scalar_specific[grid_enum::nscalars];
+#endif  // SCALAR
+
+  /// Default constructor, should init everything to zero
+  Primitive() = default;
+  /// Manual constructor, mostly used for testing and doesn't init all members
+  Primitive(Real const in_density, Vector const &in_velocity, Real const in_pressure,
```
helenarichie: Does `in_` refer to "init"? If so, `init_` might be clearer, this confused me at first.

src/utils/basic_structs.h
```diff
@@ -0,0 +1,94 @@
...
+namespace hydro_utilities
+{
+// =====================================================================================================================
+// Here are some basic structs that can be used in various places when needed
+// =====================================================================================================================
+/*!
+ * \brief A data only struct that contains the Real members x, y, and z for usage as a vector
+ *
+ */
+struct Vector {
```
helenarichie: I agree that it's confusing how this name conflicts with `std::vector` and also my brain initially wondered if it had anything to do with `cuda_utilities::DeviceVector`. Would `hydro_utilities::ComponentVector` or something like that be clearer?

src/utils/basic_structs.h
```diff
@@ -0,0 +1,94 @@
...
+  Vector magnetic;
+#endif  // MHD
+
+#ifdef DE
+  /// The specific thermal energy in the gas
+  Real gas_energy_specific;
+#endif  // DE
+
+#ifdef SCALAR
+  Real scalar_specific[grid_enum::nscalars];
```
helenarichie, line 114: I might be misunderstanding something, but I'm not sure that the `_specific` suffix makes sense for every scalar that this could refer to.

src/reconstruction/reconstruction_internals.h
```diff
@@ -0,0 +1,937 @@
...
+      loaded_data.magnetic.x = magnetic_centered.z;
+      loaded_data.magnetic.y = magnetic_centered.x;
+      loaded_data.magnetic.z = magnetic_centered.y;
+      break;
+  }
+#endif  // MHD
+
+// Load pressure accounting for duel energy if enabled
+#ifdef DE  // DE
+  Real const E          = dev_conserved[grid_enum::Energy * n_cells + id];
```
helenarichie: I think this variable should be called `energy` according to the style guide

src/reconstruction/reconstruction_internals.h
```diff
@@ -0,0 +1,937 @@
...
+      break;
+    case grid_enum::momentum_z:
+      loaded_data.magnetic.x = magnetic_centered.z;
+      loaded_data.magnetic.y = magnetic_centered.x;
+      loaded_data.magnetic.z = magnetic_centered.y;
+      break;
+  }
+#endif  // MHD
+
+// Load pressure accounting for duel energy if enabled
```
helenarichie: typo (haha)

src/reconstruction/reconstruction_internals.h
```diff
@@ -0,0 +1,937 @@
...
+#ifdef DE  // DE
+  Real const E          = dev_conserved[grid_enum::Energy * n_cells + id];
+  Real const gas_energy = dev_conserved[grid_enum::GasEnergy * n_cells + id];
+
+  Real E_non_thermal = hydro_utilities::Calc_Kinetic_Energy_From_Velocity(
```
helenarichie: Style guide => `energy_non_thermal`

src/utils/basic_structs.h
helenarichie, line 1: Overall, I really like this. I think it makes the code where it's implemented quite clear and readable.

src/reconstruction/reconstruction_internals.h
```diff
@@ -0,0 +1,937 @@
...
+ * \param[in] sound_speed The speed of sound
+ * \param[in] sound_speed_squared The speed of sound squared
+ * \param[in] gamma The adiabatic index
+ * \return Characteristic
+ */
+Characteristic __device__ __inline__ Primitive_To_Characteristic(hydro_utilities::Primitive const &primitive,
+                                                                 hydro_utilities::Primitive const &primitive_slope,
+                                                                 EigenVecs const &eigen, Real const &sound_speed,
+                                                                 Real const &sound_speed_squared, Real const &gamma)
+{
```
helenarichie, line 397: I think it's not clear what `a0`, `a1`, etc. are. I know they're associated with the `Characteristic` struct, but I don't actually know what they are (I'm not super familiar with all of the numerical hydro stuff going on here). But if they refer to the conserved quantities, I think it'd be helpful to have a comment saying which conserved variable each `a` refers to/why we're calling them e.g. `a0` instead of `density`. Or, if they're not the conserved quantities, maybe you could say what they are (like if they're term in an equation, you could say which equation)

src/reconstruction/reconstruction_internals.h
```diff
@@ -0,0 +1,937 @@
...
+
+// =====================================================================================================================
+/*!
+ * \brief Compute the Van Lear slope from the left and right slopes
+ *
+ * \param[in] left_slope The left slope
+ * \param[in] right_slope The right slope
+ * \return hydro_utilities::Primitive The Van Leer slope
+ */
+hydro_utilities::Primitive __device__ __host__ __inline__ Van_Leer_Slope(hydro_utilities::Primitive const &left_slope,
```
helenarichie: Should this be called `Compute_Van_Leer_Slope` to be consistent with the naming of `Compute_Slope`? Or, maybe even clearer, they could be `Van_Leer_Slope` and `Simple_Slope`?

src/reconstruction/reconstruction_internals.h
helenarichie, line 1: Overall, I think this looks great. I think it's quite easy to follow even though I'm not deeply familiar with the numerical hydro stuff going on. That said, know that I did not double-check all of the math in here...

src/riemann_solvers/hlld_cuda.h
```diff
@@ -124,25 +112,25 @@ struct Speeds {
  * \param o1 Direction parameter
  * \param o2 Direction parameter
  * \param o3 Direction parameter
- * \return mhd::internal::State The loaded state
+ * \return reconstruction::InterfaceState The loaded state
  */
-__device__ __host__ mhd::internal::State loadState(Real const *interfaceArr, Real const &magneticX, Real const &gamma,
-                                                   int const &threadId, int const &n_cells, int const &o1,
-                                                   int const &o2, int const &o3);
+__device__ __host__ reconstruction::InterfaceState loadState(Real const *interfaceArr, Real const &magneticX,
```
helenarichie, line 117: Again, while this stuff is being refactored, these function names could be changed to follow the style guide. Or, we could just leave them!

helenarichie (commented):
Okay, I had a look through everything, but focused on `basic_structs.h` and `reconstruction_internals.h`. I think this looks really great! I find that these changes improve readability quite a bit. I mostly had comments about naming/comments to improve clarity. You can take them or leave them!

src/utils/basic_structs.h
bcaddy: it's just `in` as in input/output. It's just to avoid name conflict with the member variables of the same name. Want me to change it to `input` or `init`? I think either makes reasonable sense.

src/utils/mhd_utilities.h
bcaddy: agreed. It predates our naming convention. Changing it would be great but I have limited time left and want to prioritize some other stuff.

src/utils/basic_structs.h
bcaddy: Maybe? "Component" shows up in other places too. Maybe something like `xyzVector`? I'd go for `Vector3D` but that might imply a 3D `std::vector` or `DeviceVector`.

src/utils/basic_structs.h
bcaddy: Physically, probably not. But in most places in the reconstructors it is what we do and all the scalars are treated the same.

src/reconstruction/reconstruction_internals.h
bcaddy: Good catch, fixed it.

src/reconstruction/reconstruction_internals.h
bcaddy: Nope, I'm making the energies fight with pistols at dawn. (fixed it)

src/reconstruction/reconstruction_internals.h
bcaddy: Fixed

src/reconstruction/reconstruction_internals.h
bcaddy: They're the characteristic variables, the primitive variables projected into a vector space where the PDE becomes an ODE. `a_1`, `a_2` etc are the standard notation for them mathematically. I agree that they're hard to reason about and that's a major part of my complaints with characteristic based methods. The `Primitive_To_Characteristic` function has the projections from primitive to characteristic variables in it.

src/reconstruction/reconstruction_internals.h
bcaddy: Yep, changed it to `Compute_Van_Leer_Slope`.

src/riemann_solvers/hlld_cuda.h
bcaddy: I think they get changed in the next PR in this series.

src/utils/basic_structs.h
helenarichie: Got it. If it's consistent with other places in the code then I think it makes sense to leave it.

src/reconstruction/reconstruction_internals.h
helenarichie: 😂

src/riemann_solvers/hlld_cuda.h
helenarichie: Fantastic!

src/utils/basic_structs.h
helenarichie: I think `xyzVector` might be the clearest, but I'm curious to hear if anyone else has any opinions

src/utils/basic_structs.h
helenarichie: Got it! Then I think it's fine as is, but a clarifying that explains what the prefix means would be helpful.

src/reconstruction/reconstruction_internals.h
helenarichie: Got it! Then I think a comment expanding on what they are in line 81 would clear this up for non-experts.

src/utils/basic_structs.h
bcaddy: I added it, how does it look now?

src/reconstruction/reconstruction_internals.h
bcaddy: I added a clarification to the struct doxygen comment, better now?

helenarichie: Beautiful

src/utils/basic_structs.h
helenarichie: Yes, thank you!

src/utils/basic_structs.h
bcaddy: changed to `VectorXYZ` per an offline discussion on the naming.

src/utils/basic_structs.h
```diff
@@ -0,0 +1,95 @@
...
+// =====================================================================================================================
+// Here are some basic structs that can be used in various places when needed
+// =====================================================================================================================
+/*!
+ * \brief A data only struct that contains the Real members x, y, and z for usage as a vector
+ *
+ */
+struct VectorXYZ {
+  Real x, y, z;
+};
```
mabruzzo: I haven't looked deeply at this PR (I can, if you want me to --  although I suspect you just want to get done at this point), but I wanted to share a quick thought.

In PR #386, I introduce a class that serves a similar purpose to this struct. I just wanted to share what my class looks like on the off-chance you would want to make use of it here too. But, please don't feel pressured to change anything (especially since it would probably trigger a number of cascading changes on your end). Given the added functionality of my class, I think it would be perfectly fine to have your struct and my functionality coexist. 

The main difference is that my class provides has some additional machinery for supporting index access. For an instance `vec`, the following pairs of expressions are equivalent: ( `vec[0]`, `vec.x()` ); ( `vec[1]`, `vec.y()` ); ( `vec[2]`, `vec.z()` ). While the functionality is a little ugly, the array-access is extremely useful for implementing logic related to positions in for-loops.

```suggestion
struct VectorXYZ {
  /* Tracks the values held by the class. To ensure the class is an aggregate, it needs
   * to be public. With that said, it should be treated as an implementation detail */
  Real arr_[3];

  /* To ensure this class is an aggregate, constructors are implicitly defined.
   * The destructor & move/copy assignment operations are also implicitly defined */

  __device__ __host__ T& operator[](std::size_t i) {return arr_[i];}
  __device__ __host__ const T& operator[](std::size_t i) const {return arr_[i];}
  __device__ __host__ T* data() {return arr_; }

  __device__ __host__ Real& x() noexcept  {return arr_[0];}
  __device__ __host__ const Real& x() const noexcept {return arr_[0];}
  __device__ __host__ Real& y() noexcept  {return arr_[1];}
  __device__ __host__ const Real& y() const noexcept {return arr_[1];}
  __device__ __host__ Real& z() noexcept  {return arr_[2];}
  __device__ __host__ const Real& z() const noexcept {return arr_[2];}
};
```

Again, I'm not really suggesting you make this change. I'm mostly just making you aware that a similar class will exist.

bcaddy: I agree that this is better and I'll add it. Thank you!

bcaddy: Added it and I made you co-author on the commit.

src/utils/basic_structs.h
evaneschneider: I think this is worth revisiting... while I appreciate that in *most* places this does end up being the specific version of the gas energy and the scalar, I think it is maybe adding "_specific" is worse than ambiguity given that in a few places (including an entire Riemann solver) it is explicitly NOT the specific version.

evaneschneider (approved):
I'm going to approve this to get the PR line moving, but I do think we should revert to just "gas_energy" and "scalar" in the structs, to avoid confusion down the line.

bcaddy:
I'm happy to make this change. I do want to point out that except for the single Riemann solver that uses the bulk quantities, in every other place used it is the specific version of the gas energy and scalar which is why I chose those names. Do you still want me to change it?

evaneschneider:
Yes, we had a discussion about it and everyone agreed that unless it is actually true everywhere, it's better to leave the name more ambiguous and comment every time the variable is first used.

evaneschneider:
As I believe Helena pointed out, it is also true that in the future there may be additional versions of scalars where it makes more sense not to use the specific version, but we don't necessarily to want to add a whole additional variable to the struct, when part of the point is to streamline/abstract things.

bcaddy:
Fair enough, I'll remove it.

bcaddy:
Done, see PR #397

Issue #372, opened 2024-02-06
Default density_floor warning does not trigger if density_floor absent from parameter file.


brantr:
In global.cpp, there is a warning that looks intended to alert the user that they are using the default `density_floor=0`.  However, as written, this will only trigger if `density_floor` is defined in the parameter file:

```
#ifdef DENSITY_FLOOR
  } else if (strcmp(name, "density_floor") == 0) {
    parms->density_floor = atof(value);
    if (parms->density_floor == 0) {
      chprintf(
          "WARNING: density floor is set to its default value (zero)! It can be set to a different value in the input "
          "parameter file.\n");
    }
#endif
```

If `density_floor` doesn't appear in the parameter file, the` else if() `doesn't get executed.

I'm checking to see whether this was intended (warning the user they have set the density floor to zero), or whether a general warning about using `density_floor=0` was intended (regardless of the declaration of `density_floor` in the parameter file).

brantr:
I have verified that setting `density_floor=0` in the parameter file triggers the warning.

evaneschneider:
It should be the case that not setting the density floor parameter in the input file still triggers the warning, since it is given a default value of zero in the parameter struct header where it is declared. If not, we should fix it. The intended behavior is to warn the user if they have compiled with the density floor flag but haven’t set their desired floor in the input file. If this is unclear, we can work on the wording in the warning.

--
Evan Schneider
Assistant Professor of Physics and Astronomy
University of Pittsburgh
520-822-6294
pronouns: she/they
I support the Union of Pitt Faculty<https://www.pittfaculty.org/>

On Feb 6, 2024, at 5:23 PM, Brant Robertson ***@***.***> wrote:

﻿

I have verified that setting density_floor=0 in the parameter file triggers the warning.

evaneschneider:
@helenarichie Can you take a look at this?

evaneschneider:
Fixed in PR #373

PR #373, opened 2024-02-08
Patch to fix logic of floor warning messages

helenarichie:
As raised in Issue #372, the previous version of the warning messages for the temperature, density, and scalar floors only worked when the `temperature_floor`, `density_floor`, and `scalar_floor` parameters were set in the input parameter file. The intended behavior is for them to be thrown when the parameters are _not_ set or are set to zero by checking whether their values in the header are zero (the default value). I moved the check and error messages outside of `Parse_Param` after it is called. Otherwise, the warning is thrown multiple times as `Parse_Param` loops over the other parameters. Let me know if it looks okay.

Thanks @brantr for catching this!

### DIFF
--- a/src/global/global.cpp
+++ b/src/global/global.cpp
@@ -185,27 +185,27 @@ void Parse_Params(char *param_file, struct Parameters *parms, int argc, char **a
     Parse_Param(name, value, parms);
     chprintf("Override with %s=%s\n", name, value);
   }
-  #ifdef TEMPERATURE_FLOOR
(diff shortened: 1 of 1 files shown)


evaneschneider:
Looks good to me! Did you test the new behavior with/without the parameters set in the input file?

helenarichie:
I did! It throws the warning when they're set to zero/not set and does nothing when the parameters are there and set to something other than zero.

evaneschneider:
Great, I'll merge it in and close the issue.

PR #375, opened 2024-02-13
Reconstruction Kernel Fusion 2: New Loading and Conversion Utility Functions

bcaddy:
# Summary

The primary purpose of this PR is to introduce 4 new utility functions for loading grid data and converting between the primitive and conserved variables. This builds off of PR #371 and will show changes from both that PR and this PR until #371 is merged into `dev`. Most of the relevant changes are in the following files:

- hydro_utilities.h
- plmc_cuda.h and .cu
- ppmc_cuda.h and .cu

The data loading functions are both templates that take the direction as a template parameter and default to `dir = 0`. Per Issue #308 we've seen that this can improve performance noticeably. After integrating these new templated functions into PLMC and PPMC (and making them templates in the same way) I found a ~6% improvement in performance on a single V100.  ([timing data](https://github.com/cholla-hydro/cholla/files/14271116/run_timing.log))

## `hydro_utilities::Load_Cell_Conserved`

This function loads the conserved data from the `dev_conserved` array and returns an assembled `Conserved` object.

## `hydro_utilities::Load_Cell_Primitive`

This function calls `hydro_utilities::Load_Cell_Conserved` and then `hydro_utilities::Conserved_2_Primitive` to load the conserved data, convert it to primitive, and returns a `Primitive` object.

## `hydro_utilities::Conserved_2_Primitive`

Converts the conserved variables in a `Conserved` object into primitive variables in a `Primitive` object and returns it.

## `hydro_utilities::Primitive_2_Conserved`

Converts the primitive variables in a `Primitive` object into conserved variables in a `Conserved` object and returns it.

## `math_utils::Cyclic_Permute_Once` and `math_utils::Cyclic_Permute_Twice`

These two functions take a `hydro_utilities::Vector` object and cyclically permute its members once or twice. It's used in the `Load_Cell_Conserved` function and I split it out on its own for clarity and because I thought it might be more generally useful. I tried to think of a single function that would do this more elegantly but I couldn't figure out a way, if you have suggestions I would be happy to hear them.

## Other

- Added a constructor for `hydro_utilities::Conserved`
- Fixed a bug in `reconstruction::Kind`, it had `PLMC` twice when one of them should have been `PPMC`.

### DIFF
--- a/src/integrators/VL_1D_cuda.cu
+++ b/src/integrators/VL_1D_cuda.cu
@@ -104,7 +104,7 @@ void VL_Algorithm_1D_CUDA(Real *d_conserved, int nx, int x_off, int n_ghost, Rea
                      gama, 0, n_fields);
   #endif
   #ifdef PPMC
-  hipLaunchKernelGGL(PPMC_VL, dimGrid, dimBlock, 0, 0, dev_conserved_half, Q_Lx, Q_Rx, nx, ny, nz, gama, 0);
+  hipLaunchKernelGGL(PPMC_VL<0>, dimGrid, dimBlock, 0, 0, dev_conserved_half, Q_Lx, Q_Rx, nx, ny, nz, gama);
   #endif
   GPU_Error_Check();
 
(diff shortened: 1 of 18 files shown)


evaneschneider (approved):
This all looks good.

PR #376, opened 2024-02-16
Remove Monotonization in PLMC

bcaddy:
This monotonization is no longer required after the bug fix in the utility functions and it too often reduces the method to first order so we're removing it. I also had to update many of the system tests for this change. See #361 for more info on the bug fixes.

To catch these kinds of issues in the future I added a system test for the Kelvin Helmholtz instability since during testing it showed the issue quite noticeably.


ojwg (approved):
Change seems very straightforward.  Software-wise, looks good.

PR #378, opened 2024-03-07
Convert `N_STEPS_LIMIT` macro to runtime argument

bcaddy:
The `N_STEPS_LIMIT` macro is now a runtime argument under the name `n_steps_limit`


bcaddy:
The wiki has been updated with this. I also added documentation for `legacy_flat_outdir`.

ojwg (changes requested):
I have a few concerns.  Now that the limits have been removed from the make.type files, shouldn't we add the removed values into the relevant saved parameter files?  Otherwise unsuspecting users may not be aware that the limits are gone.

Looking at the expression `G.H.n_step == P.n_steps_limit and P.n_steps_limit > 0`, there could be a use case where someone is restarting a sim at a certain time where the G.H.n_step starts out at a higher value than a newly specified `n_steps_limit` (granted, this is a user mistake).  In this case the limit check will never trigger.  I propose something like `G.H.n_step >= P.n_steps_limit and  P.n_steps_limit > 0`.

bcaddy:
> I have a few concerns. Now that the limits have been removed from the make.type files, shouldn't we add the removed values into the relevant saved parameter files? Otherwise unsuspecting users may not be aware that the limits are gone.

Evan and I discussed that. The only case where it matters is the FOM and PR #380 adds FOM parameter files so we can update those with the limits.


> Looking at the expression G.H.n_step == P.n_steps_limit and P.n_steps_limit > 0, there could be a use case where someone is restarting a sim at a certain time where the G.H.n_step starts out at a higher value than a newly specified n_steps_limit (granted, this is a user mistake). In this case the limit check will never trigger. I propose something like G.H.n_step >= P.n_steps_limit and  P.n_steps_limit > 0.

Great point, I'll update it now.

ojwg (approved):
looks good

PR #380, opened 2024-03-08
CAAR FOM Case

reubendb:
Minor fixes to get CAAR FOM case to run. Also include example input files and jobscript stub.


src/io/io.cpp
bcaddy, line 1: What's the purpose of these changes?

src/particles/io_particles.cpp
bcaddy, line 1: What's the purpose of moving this line out of the ifdef?

bcaddy (changes requested):

src/particles/io_particles.cpp
brantr: I suspect @reubendb was using `TILED_INITIAL_CONDITIONS`, so `nfile` would not be defined when it's called by `Load_Particles_Data_HDF5(file_id, nfile, P); `later in the function.  I think it should actually go outside the `MPI_CHOLLA` `ifdef` into the main body of the function.

src/io/io.cpp
brantr: On Frontier, when compiling, the following error is thrown:

```
src/io/io.cpp:2275:3: error: no matching function for call to 'sprintf'
  sprintf(filename, "%sics_%dMpc_%d.h5", P.indir, (int)P.tile_length / 1000,
  ^~~~~~~
/usr/include/stdio.h:334:12: note: candidate function not viable: no known conversion from 'std::string' (aka 'basic_string<char>') to 'char *__restrict' for 1st argument
extern int sprintf (char *__restrict __s,
           ^
```

So either something like` filename.assign()` or another char string needs to be used.

src/particles/io_particles.cpp
brantr: Indeed, when compiling on Frontier a FOM run we get:

```
src/particles/io_particles.cpp:59:37: error: use of undeclared identifier 'nfile'
  Load_Particles_Data_HDF5(file_id, nfile, P);
                                    ^
```
I'd suggest moving the line

  const int nfile              = P->nfile;  // output step you want to read from

outside all of the ifdefs, immediately after the HDF5 check.

src/io/io.cpp
bcaddy: In that case, I think that rather than creating a static `char` array we should just use the 'c_str()` or `data()` methods of  `std::string`. It's safer and doesn't run into the issue of having being too short.

src/particles/io_particles.cpp
bcaddy: I agree, moving it out of the `MPI_CHOLLA` check is the right move.

src/io/io.cpp
reubendb: `sprintf()` needs to write to a char array. Both `c_str` and `data` method just get the C string and pointer of `std::string`, respectively, and my understanding is one cannot write to them. 
Overflow is a valid concern; not sure what to do with `sprintf()` other than making larger buffer (suggestions?).

src/particles/io_particles.cpp
reubendb: It was originally there, outside of all ifdef.  Someone moved them inside one of the ifdef. Not sure what the motivation was; I suspected possibly to avoid un-used declared variable so  I just tried to follow it (i.e. minimal change that fix the problem).

evaneschneider: Leaving it inside of the MPI ifdef seems safest if we don't know why it was moved there.

src/io/io.cpp
evaneschneider: Given the implementation @reubendb wrote works, I think this PR is fine as-is, and we can optimize it later in a separate PR.

bcaddy (approved):

src/particles/io_particles.cpp
brantr: If the `nfile` definition is moved outside the MPI ifdef, I think the code will not compile without `MPI_CHOLLA` set.

PR #386, opened 2024-03-26
Feedback overhaul

mabruzzo:
This should be reviewed after #418 and #420 are merged

---------

This overhauls stellar feedback and makes a few tweaks to the gravity solver to make it possible to run with particles so that the particles feel self-gravity and the fluid only feels a static potential.


mabruzzo:
TODO: fix docstring of `Apply_Energy_Momentum_Deposition`

mabruzzo:
pre-commit.ci autofix

mabruzzo:
pre-commit.ci autofix

.github/workflows/compilation_checks.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         make-type: [hydro, gravity, disk, particles, cosmology, mhd, dust, cooling]
         container:
           #- {name: "CUDA", link: "docker://chollahydro/cholla:cuda_github"}
-          - {name: "HIP", link: "docker://chollahydro/cholla:rocm_github"}
+          - {name: "HIP", link: "ghcr.io/cholla-hydro/cholla-rocm:sha-a08bd2c"}
```
evaneschneider, line 29: Is this intentional? For some reason I thought this had been updated elsewhere.

builds/make.type.disk
evaneschneider, line 1: Could you add a brief comment to the top line of this makefile (and the starblast) just stating the basic purpose (like in the hydro and mhd files)?

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,323 @@
...
+
+where
+-  {math}`\Phi_{\rm dm}` corresponds to the dark matter halo  (always a static potential)
+-  {math}`\Phi_{\rm stars,old}` corresponds to older stellar populations in the disk (always a static potential)
+- {math}`\Phi_{\rm stars,young}` corresponds to the younger stellar populations (computed with self-gravity from particles)
+- {math}`\Phi_{\rm gas}` corresponds to the younger stellar populations (computed with self-gravity from particles)
+
+### Primer on Radiative Cooling
+
+At the time of writing, simulations must always be initialized with CIE cooling (cooling shuts off below 1e4 K) in order to prevent immediate of the gas disk. The current strategy for using other cooling mechanisms. After running the simulation long enough to start driving turbulence in the gas disk (commonly a few 10s of Myr), you can then restart the simulation with a different cooling recipe.
```
evaneschneider: typo: "immediate **collapse** of the..."

evaneschneider: Also the following sentence is not complete.

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,323 @@
...
+
+### Primer on Feedback and Star Formation:
+
+In a simulation without particles, we feedback is implemented with a CGOLs like mechanism.
```
evaneschneider: This option is not actually included in this version of cholla, right? My recollection is that the CGOLS feedback code has only ever existed in branches that have not been merged into this repo, in which case, maybe we should not mention it here?

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,323 @@
...
+:::
+
+Consider the parameterization of the potential of a Miyamoto-Nagai potential (this will be important [again](#analytic-self-gravity-estimate), shortly)
+
+:::{math}
+\Phi_{\rm MN}(R,z; M,a,b)\equiv \frac{-G M}{\sqrt{R^2 + (a + \sqrt{z^2 + b^2})^2}}.
+:::
+
+At the time of writing, we always define {math}`\Phi_{\rm stars,old}(R,z)=\Phi_{\rm MN}(R,z; M_{\rm stars}, R_{\rm stars}, z_{\rm stars})`, where {math}`R` & {math}`z` are cylindrical coordinates, {math}`R_{\rm stars}` & {math}`z_{\rm stars}` are the scale radius/height, and {math}`M_{\rm stars}` is the mass of the disk.
+
```
evaneschneider, line 101: Would it be possible in this section to give a brief description of how these parameters are set? I.e. are they hardcoded in the disk model, are they runtime parameters...?

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,323 @@
...
+As noted in the [Schneider & Robertson 2018](https://ui.adsabs.harvard.edu/abs/2018ApJ...860..135S) , "in setting the halo gas distribution the potential is taken to be spherically symmetric." In other words, the halo gas profile uses a potential called {math}`\Phi_{\rm sph}(r) = \Phi_{\rm stars}(R=0,z=r) + \Phi_{\rm halo}(r)` (it **NEVER** includes contributions from self-gravity). We use an isentropic profile:
+
+:::{math}
+
+\rho_{\rm h}(r) = \rho_{\rm h}(r_0)\left[ 1+(\gamma - 1) \frac{\Phi_{\rm sph}(r) - \Phi_{\rm sph}(r_0)}{c_{s,{\rm h}}^2(r_0)}\right]^{1/(\gamma-1)}.
+
+:::
+
+In the above equation
+- Consequently, we define . At the time of writing, this **never** includes contributions from self-gravity. (this definition of self gravity means that we can **NEVER** achieve perfect )
```
evaneschneider: Couple missing phrases.

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,323 @@
...
+:::
+
+Essentially we can use this to incremental rate of star-formation {math}`d{\rm SFR}` in the disk between {math}`r_{\rm cyl}` and {math}`(r_{\rm cyl} + dr_{\rm cyl}`). If we consider some duration of time (in practice set by the `t_out` runtime parameter), you can work an expected number of stars formed between  {math}`r_{\rm cyl}` and {math}`(r_{\rm cyl} + dr_{\rm cyl}`) during that duration. It's straight-forward to convert this function into a PDF.
+
+:::{note}
+I can definitely elaborate more -- I found the relevant notes on this topic
+:::
+
+At startup we sample this PDF to determine the {math}`r_{\rm cyl}` at which all particles are expected to form. We create star-particles at these radii and distribute the "turn-on times" from a time a little before we start the simulation until the time specified by the `tout` parameter. We use Poisson sampling to distribute these "turn-on times," to target a particular SFR. 
+
```
evaneschneider, line 366: Similar comment here about how the SFR is set. Is this a runtime parameter, or hardcoded somewhere?

docs/sphinx/Physics/Feedback.md
```diff
@@ -1,9 +1,168 @@
...
+
+:::{note}
+While this "sequencing" solution isn't very "physical," it only makes a practical difference when stencils overlap.
+If it is coming into play frequently in a given simulation, we may want to consider alternatives, like shorter timesteps or subcycling (after all, the probability that 2 nearby star clusters would have a supernova at exactly the same time is extremely low)
+:::
+
+The main alternative that we could revisit is implementing the method prescriptions as atomic operations. If we want to adopt this solution in the future, there are a number of important considerations that need to be addressed in the future. These details are highlighted under the collapsible tag
+
+
+#### ASIDE: Important considerations for atomic "conflict-resolution"
```
evaneschneider, line 67: Is this supposed to be collapsible?

docs/sphinx/Physics/DiskModel.md
mabruzzo: Fixed

docs/sphinx/Physics/DiskModel.md
mabruzzo: Yes. I think so. (It wasn't clear to me back when I first wrote these docs)

docs/sphinx/Physics/Feedback.md
mabruzzo: Good catch. The section was originally collapsible on the GitHub wiki.

With the switch to sphinx, we lost the ability to do that (so I just discuss things in the next subsection). We could probably find a sphinx extension to do that (but it may not be worth doing).

docs/sphinx/Physics/DiskModel.md
mabruzzo: Its hardcoded. I added a brief description

docs/sphinx/Physics/DiskModel.md
mabruzzo: Its hardcoded (I added a description)

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+:::{math}
+
+\Phi_{\rm tot} = \Phi_{\rm dm} + \Phi_{\rm stars,old} + \Phi_{\rm stars,young} + \Phi_{\rm gas}
+:::
+
+where
+-  {math}`\Phi_{\rm dm}` corresponds to the dark matter halo  (always a static potential)
+-  {math}`\Phi_{\rm stars,old}` corresponds to older stellar populations in the disk (always a static potential)
+- {math}`\Phi_{\rm stars,young}` corresponds to the younger stellar populations (computed with self-gravity from particles)
+- {math}`\Phi_{\rm gas}` corresponds to the younger stellar populations (computed with self-gravity from particles)
```
evaneschneider:
```suggestion
- {math}`\Phi_{\rm gas}` corresponds to the gas disk (computed with self-gravity from particles)
```

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+This configuration has not been tested for quite a while in the dev branch.
+:::
+
+:::{caution}
+At the time of writing, the logic modelling the static potential is duplicated between the initial conditions and the source-term calculation.
+It is the user's responsibility to ensure that consistent models of the potential are used in both parts of the code.
+:::
+
+2. Normal Self Gravity
+  - use the self-gravity module to model full gravitational potential while updating timesteps.
```
evaneschneider:
```suggestion
  - use the self-gravity module to model the full gravitational potential while updating timesteps.
```

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+All configurations of the idealized Galaxy Simulations:
+- currently assume that you are using a static Gravitational Potential produced by 2 components: (i) component contributed by older stellar populations {math}`\Phi_{\rm stars-old}` and (ii) a component contributed by dark matter in the halo, {math}`\Phi_{\rm halo}`.
+- use these potentials for initializations and for evolving the simulation over a timestep.
+
+:::{important}
+[As we noted above](#configurations), when using the standalone Static-Gravity module, the logic implementing the static potential is duplicated between the initial conditions and the source-term calculations; it's the user's responsibility to ensure logic remains consistent in both parts of the code. 
+This is **NOT** a concern when using the Self-Gravity module
+:::
+
+Consider the parameterization of the potential of a Miyamoto-Nagai potential (this will be important [again](#analytic-self-gravity-estimate), shortly)
```
evaneschneider:
```suggestion
Consider the parameterization of a Miyamoto-Nagai potential (this will be important [again](#analytic-self-gravity-estimate), shortly)
```

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+At the time of writing, we always define {math}`\Phi_{\rm stars,old}(R,z)=\Phi_{\rm MN}(R,z; M_{\rm stars}, R_{\rm stars}, z_{\rm stars})`, where {math}`R` & {math}`z` are cylindrical coordinates, {math}`R_{\rm stars}` & {math}`z_{\rm stars}` are the scale radius/height, and {math}`M_{\rm stars}` is the mass of the disk.
+Currently, these values are hardcoded where we initialize the `ClusteredDiskGalaxy` struct that holds the Milky Way properties (in {repository-file}`src/model/particles/disk_galaxy.cu`).
+
+At the time of writing, {math}`\Phi_{\rm dm}` is always assumed to be an NFW profile, or
+
+:::{math}
+\Phi_{\rm dm}(r) = \frac{-G M_{\rm vir}}{r[\ln(1+c) - c/(1+c)]}\ \ln \left(1 +\frac{r}{R_{\rm dm}}\right),
+:::
+
+where {math}`r=\sqrt{R^2+z^2}` is radius in spherical coordinates, {math}`M_{\rm vir}` is the dark matter mass, {math}`c` is halo concentration, and {math}`R_{\rm dm}` is the scale length of the halo.
```
evaneschneider, line 108: Could you also add where these values are set?

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+- more importantly: {math}`\hat{\Phi}_{\rm gas,disk}(R,z)` includes the contributions to the gravitational potential from material outside of the simulation domain (if that sounds "wrong" to you, that's becomes most of your physical intuition comes from spherical symmetry).
+
+## Gas Initial Conditions
+
+The ``"Disk_3D"`` and ``"Disk_3D_particles"`` initial conditions will both initialize the gas disk. Both cases initialize an isothermal gas disk within a halo. The latter also subsequently initializes particles.
+
+Cholla offers machinery to setup an idealized galaxy simulation. Specifically, it initializes properties a gas disk inside of gas halo. The conditions broadly match the description provided by  [Schneider & Robertson 2018](https://ui.adsabs.harvard.edu/abs/2018ApJ...860..135S) (there are some extensions when using self-gravity).
+
+Ideally, we our initial conditions would be in hydrostatic equilibrium (i.e. the properties of a simulation would not change if we hit "go"). For a handful of reasons, we can't actually achieve initial conditions a handful of reasons (related to the gas in the halo and the boundary conditions), but we pick ICs that are stable.
```
evaneschneider:
```suggestion
Ideally, our initial conditions would be in hydrostatic equilibrium (i.e. the properties of a simulation would not change if we hit "go"). For a handful of reasons, we can't actually achieve perfectly hydrostatic initial conditions (related to the gas in the halo and the boundary conditions), but we pick ICs that are stable.
```

docs/sphinx/Physics/DiskModel.md
```diff
@@ -0,0 +1,335 @@
...
+
+:::{math}
+
+\Sigma_{\rm SFR}(R) = a * \Sigma_{{\rm gas},0}^{k_s} * \exp(-r_{\rm cyl} / R_{\rm gas-scale-length}).
+:::
+
+Essentially we can use this to incremental rate of star-formation {math}`d{\rm SFR}` in the disk between {math}`r_{\rm cyl}` and {math}`(r_{\rm cyl} + dr_{\rm cyl}`). If we consider some duration of time (in practice set by the `t_out` runtime parameter), you can work an expected number of stars formed between  {math}`r_{\rm cyl}` and {math}`(r_{\rm cyl} + dr_{\rm cyl}`) during that duration. It's straight-forward to convert this function into a PDF.
+
+:::{note}
+I can definitely elaborate more -- I found the relevant notes on this topic
```
evaneschneider: Given the questions that arose the other day, I think that might be a good idea.

docs/sphinx/Physics/Feedback.md
```diff
@@ -1,9 +1,168 @@
...
+1. **Thermal Energy, Momentum, and Mass injection:** in practice this isn't any more complex than the last case (the same strategy applies)
+2. **Dynamically adjusting the prescription based on local conditions:** theoretically you can determine the kind of prescription each particle will use first, and then apply the same procedure that was used in the last 2 cases. In practice, this may not be sensible (e.g. if deciding between resolved vs unresolved)
+3. **Prescriptions that can average values within stencils in addition to injections:** I don't think there is a robust, straight-forward way to make this work with atomics (that also scales to arbitrarily complex scenarios)
+
+(SNe-Prescription-Descriptions)=
+## Prescription
+
+This section describe's the actual prescriptions. First we offer some broad background context. Then, we offer a high-level description of the available prescriptions. Subsequently, we provide additional detail about the stencils that are actually used and the magnitudes of the injected quantities.
```
evaneschneider:
```suggestion
This section describes the actual prescriptions. First we offer some broad background context. Then, we offer a high-level description of the available prescriptions. Subsequently, we provide additional detail about the stencils that are actually used and the magnitudes of the injected quantities.
```

docs/sphinx/Physics/Feedback.md
```diff
@@ -1,9 +1,168 @@
...
+    Unresolved feedback uses a legacy 27-cell stencil based on CiC Interpolation (i.e. the stencil is effectively a cube)
+  - **IMPORTANT:** please see the description of the momentum stencil before you pick this choice.
+- {par:param}`feedback.sn_model` = `"legacyAlt"`
+  - kind: hybrid
+  - stencil: - stencil: resolved feedback uses standard 8-cell Cloud-in-Cell (it's **exactly** like `"resolvedCiC"`). Unresolved feedback uses a spherical 27-cell stencil, with supersampling to calculate overlap
+  - **EXPERIMENTAL:** (this is probably a better choice that `"legacy"` since the momentum-stencil is better defined)
+
+:::{todo}
+We may want to drop a few unneeded models, and rename some options
+:::
```
evaneschneider, line 128: Agreed. At the very least I think we can drop 'legacy' altogether, and I'd suggest renaming legacyAlt to something more descriptive like 'hybrid'. Your tests demonstrated that legacy alt worked better than legacy, right?

docs/sphinx/Physics/Gravity.md
```diff
@@ -16,33 +34,125 @@ Static gravity is activated using the ```STATIC_GRAV``` macro, and is used in se
...
+In addition to static gravity, Cholla has an FFT-based self gravity solver.
+Only one or the other may be used.
+The self-gravity solver is turned on with the ``GRAVITY`` macro in the makefile.
+The default behavior in the [make.type.gravity](https://github.com/cholla-hydro/cholla/blob/dev/builds/make.type.gravity) build (and builds that depend on it) is also to turn on the ``GRAVITY\_GPU`` macro, which ensures that gravity fields reside on the GPU (required for gpu-based MPI communications), and the ```PARIS``` macro, which specifies that the Poisson solve will be carried out on the GPU by the cuFFT or rocFFT libraries.
```
evaneschneider:
```suggestion
The default behavior in the [make.type.gravity](https://github.com/cholla-hydro/cholla/blob/dev/builds/make.type.gravity) build (and builds that depend on it) is also to turn on the ``GRAVITY_GPU`` macro, which ensures that gravity fields reside on the GPU (required for gpu-based MPI communications), and the ```PARIS``` macro, which specifies that the Poisson solve will be carried out on the GPU by the cuFFT or rocFFT libraries.
```

docs/sphinx/Physics/Gravity.md
```diff
@@ -16,33 +34,125 @@ Static gravity is activated using the ```STATIC_GRAV``` macro, and is used in se
...
+Instead of requiring periodic boundaries, this variant requires the following boundary conditions:
+    1. {math}`{\bf g}({\bf x}_{\rm bound}) = {\bf 0}`, or equivalently {math}`{\bf \nabla} \phi \rvert_{{\bf x}_{\rm bound}} = {\bf 0}`
+    2. {math}`\rho({\bf x}_{\rm bound}) = 0`, or equivalently {math}`\nabla^2\phi\rvert_{{\bf x}_{\rm bound}} = 0`
+    While this modified-scheme no longer requires periodic boundary conditions, it clearly is **NOT** the full solution (only pathological scenarios can satisfy condition 1).
+
+2) **A static estimate for the gravitational potential produced by {math}`\rho_{\rm tot}`**: In more detail, we actually need a static estimate for the density-potential pair
+   - Added context: in case you aren't familiar with the concept, every valid gravitational potential profile is associated with a unique density profile. Thus we describe the pair of profiles as a "density-potential" pair [^density-potential-pair]
+   - Let's call refer denote these quantities as {math}`\rho_{\rm estimate}({\bf x})` and {math}`\phi_{\rm estimate}({\bf x})`
```
evaneschneider:
```suggestion
   - Let's call these quantities {math}`\rho_{\rm estimate}({\bf x})` and {math}`\phi_{\rm estimate}({\bf x})`
```

docs/sphinx/Physics/Gravity.md
```diff
@@ -16,33 +34,125 @@ Static gravity is activated using the ```STATIC_GRAV``` macro, and is used in se
...
+If {math}`{\bf \nabla} \phi_{\rm estimate}` isn't a good estimate for {math}`{\bf \nabla} \phi_{\rm tot}` at the boundaries **OR**  {math}`\rho_{\rm estimate}` isn't a good estimate for {math}`\rho_{\rm tot}`, then the artifacts and inaccuracies will be introduced into the solution
+:::
 
-```GRAVITY_ANALYTIC_COMP```: Add an analytic component to the gravitational potential. As of 10-27-2023, this is hard-coded to a Milky Way galaxy model in the function `Setup_Analytic_Potential` from [gravity_functions.cpp](https://github.com/cholla-hydro/cholla/blob/dev/src/gravity/gravity_functions.cpp). 
 
-```PARIS_3PT```: Use a 3-point gradient for the divergence operator approximation in Paris (default behavior is to use a spectral method)
+## Support for Problem-types with Non-Periodic Boundaries
 
-```PARIS_5PT```: Use a 5-point gradient for the divergence operator approximation in Paris
+Currently, the only scenario that Cholla supports with non-periodic boundaries is idealized Galaxy simulations without.
```
evaneschneider:
```suggestion
Currently, the only scenario that Cholla supports with non-periodic boundaries is idealized Galaxy simulations.
```

docs/sphinx/Physics/Gravity.md
```diff
@@ -16,33 +34,125 @@ Static gravity is activated using the ```STATIC_GRAV``` macro, and is used in se
...
 
-```PARIS_GALACTIC```: Use the Paris Poisson solver on a domain with analytic boundaries set to match the selected model in the DiskGalaxy class. As of 10-27-2023, this is hard-coded to a Milky Way galaxy model in the function `Compute_Gravitational_Potential` from [gravity_functions.cpp](https://github.com/cholla-hydro/cholla/blob/dev/src/gravity/gravity_functions.cpp) and in `Compute_Potential_Isolated_Boundary` from [gravity_boundaries.cpp](https://github.com/cholla-hydro/cholla/blob/dev/src/gravity/gravity_boundaries.cpp).
+At the time of writing, current density-potential pairs used in these problems currently have some **minor** "flaws:"
```
evaneschneider:
```suggestion
At the time of writing, current density-potential pairs used in these problems have some **minor** "flaws:"
```

docs/sphinx/Physics/Gravity.md
```diff
@@ -16,33 +34,125 @@ Static gravity is activated using the ```STATIC_GRAV``` macro, and is used in se
...
 
-```PARIS_GALACTIC_5PT```: Same as above but for the analytic boundary version
+[^1]: In more detail, we use the fairly standard second-order cloud-in-cell (CIC) interpolation technique to distribute the particle mass onto the field. This total density field is the primary input for this solver.
 
-```PARIS_GALACTIC_TEST```: Turn on to test whether Paris returns the same gravitational potential as the SOR solver. Doesn't work with GRAVITY_GPU, should probably be deprecated.
+[^negative-density]: If you are still struggling to wrap your head around this, it may be instructive to recall the parallels with electrostatics. In that context, Poisson's equation, with physical different constants, relates electrostatic potential and charge density (which can be negative).
 
-## Self Gravity: SOR based
-To-do: Describe the SOR solver
+[^density-potential-pair]: For the uninitiated, a given density profile always has a unique gravitational potential (modulo a constant offset in the potential).  Some simple examples of well known potential pairs include:
```
evaneschneider: Missing examples here? Or are the last two floating bullet points on the page supposed to be associated with this footnote?

docs/sphinx/Reference/param/Feedback.md
```diff
@@ -0,0 +1,57 @@
...
+
+:::{par:parameter} feedback.sn_rate
+
+:Summary: Specifies the kind of supernova rate
+:Type: {par:typefmt}`str`
+:Default: `"table"`
+
+When `"table"` (the default value) is specified, Cholla determines the rate from a table.
+`"immediate_sn"` schedules a single supernova to occur, immediately after the simulation starts.
+More details are provided ({ref}`here <general-SNe-rate>`
```
evaneschneider:
```suggestion
More details are provided {ref}`here <general-SNe-rate>`.
```

docs/sphinx/Physics/Gravity.md
mabruzzo: They are indeed. It's unfortunate that they don't render properly. I guess it's an issue related to the markdown parser (it would have worked with restructured text).

I'll rework them so that they are part of a single paragraph (and there aren't bullets).

.github/workflows/compilation_checks.yml
mabruzzo: While we generated the new docker image a while back, we haven't actually been using it (we only started needing it for this PR)

src/feedback/stencil.h
```diff
@@ -0,0 +1,937 @@
...
+  // we are intentionally using integer division to speed up the next line
+  // NOLINTNEXTLINE(bugprone-integer-division)
+  const Real l_offset = ((CellsPerDiameter % 2) == 0) ? CellsPerDiameter / 2 : 0.5 * CellsPerDiameter;
+
+  // Step 1: along each axis, identify the integer-index of the leftmost cell covered by the stencil.
+  const int leftmost_indx_x = int(pos_indU[0] - l_offset);
+  const int leftmost_indx_y = int(pos_indU[1] - l_offset);
+  const int leftmost_indx_z = int(pos_indU[2] - l_offset);
+
+  static_assert(CellsPerDiameter == 2);  // this is temporary!
```
evaneschneider: What does the temporary comment mean here?

evaneschneider:
Okay, I finally made it through the whole PR! Thanks for addressing all my previous comments. I think once the merge conflicts are sorted, this will be ready to merge.

src/feedback/stencil.h
mabruzzo: I'll remove that

mabruzzo: That was from debugging

docs/sphinx/Physics/DiskModel.md
mabruzzo: This is now addressed in PR #524

PR #390, opened 2024-04-16
Cosmology System Test & Gravity Support in `SystemTestRunner` Class 

bcaddy:
## Summary

- Added gravity support to the `SystemTestRunner` class
- Added a cosmology system test. Thank you to @bvillasen for providing the initial conditions and parameter files and helping answer my questions about this test problem.
- Added Dual Energy to the default cosmology build
- Make PPMP default for cosmo builds per offline discussion with @evaneschneider  and @bvillasen 
- Updated the `combine_hydro_particles.py` script to work with gravity files as well. Now named `combine_test_files.py`
- Fixed a few typos


evaneschneider:
Is this ready to be merged? There is still a conflict with the test data.

bcaddy:
I fixed the conflict. I'm running into a weird bug with some tests with DE turned on failing. I'll let you know when I have that resolved

bcaddy:
This is ready to merge.

The bug was due to cosmology using a different value of `DE_ETA_1` which changed the pressure in some tests. I've added handling for this to those tests.

evaneschneider:
Just so I understand, which tests exactly did you need to change? The change by several orders of magnitude in the pressure is surprising to me, but maybe it's because I don't understand what is being compared.

bcaddy:
`tALLLoadCellPrimitive` and `tALLReconstructionLoadData`. `DE_ETA_1` is very different in cosmology builds so which leads to the difference.

evaneschneider:
I'm still not understanding. Don't those tests just load in fidicuial data? So why is the data different? I understand that the eta value is different, but I'm not understanding what the test data represents.

bcaddy:
Both of the functions being tested call the "get pressure from DE" function which internally depends on the value of `DE_ETA_1` and the difference in that is enough with this fiducial data to change how it's handled.

evaneschneider:
I see, thank you.

PR #391, opened 2024-04-18
Merge PLMC and PLMP Into One & Add MHD Support to PLMP

bcaddy:
# Summary

This merges the PLMC and PLMP kernels into one single kernel. It also borrows quite a bit of the refactor from PR #382, for improved modularity, function naming, etc, just without fusing PLM into the Riemann solvers.

## PLMC/PPMP Merge

- Minor refactors/reformatting throughout now that PLMC and PLMP have been merged
- Removed PLMP kernel and associated files
- Moved most of PLMC into a device function
- Moved the PLM characteristic tracing into a device host function
- Added a switch in the PLMC function to switch between limiting in the primitive and characteristic variables
- Renamed all PLMC stuff to PLM now that the kernel does both
- Added PLMP support in PLMC (not PLM) tests
- Removed `Monotonize_Characteristic_Return_Primitive` and replaced it with the overloaded `Van_Leer_Limiter` function which works with Real, Primitive, and Characteristic structs and a subsequent call to `Characteristic_to_Primitive`.
- This also should improves PLMP for CTU/`SIMPLE` based builds by using the more sophisticated characteristic tracing in PLMC for PLMP as well.

#### Timing

PLMP is 6.7% faster than PLMC in hydro but only 4.7% faster when MHD is turned on. [run_timing.log](https://github.com/cholla-hydro/cholla/files/15068078/run_timing.log)

#### Accuracy & Stability

In hydro PLMP and PLMC are virtually identical. A suite of test plots comparing current PLMP, previous PLMP, and PLMC with both integrators can be found [here](https://github.com/cholla-hydro/cholla/files/15068175/plmp_tests.zip), it also includes the MHD PLMC vs PLMP test plots.

In MHD, PLMP is more oscillatory near shocks and gives a slightly different profile for the internal energy in the Einfeldt Strong Rarefaction. However, in more complex tests, like the blast wave, the PLMP and PLMC reconstructors perform nearly identically. 

Here's the script I used to run all the tests and make the plots. Note that it makes some assumptions about the paths to the Cholla and analysis-scripts repos. [run-plot-tests.sh.txt](https://github.com/cholla-hydro/cholla/files/15068312/run-plot-tests.sh.txt). I was on commit [05545d4](https://github.com/cholla-hydro/analysis_scripts/commit/05545d475c4ccb3252aaed7bf050809cf5e2f327) of the analysis-scripts repo.

## Other

- Standardized templates to use `uint` type for direction template arguments. It was a mix of `int` and `size_t`. Since the direction should always be 0, 1, or 2 I chose `uint` since it must be positive and `size_t` implies some kind of array indexing.
- Added `sound_speed` to the `Eigenvectors` struct. This simplifies the overall data flow quite a bit.

## Dependencies

Depends on PRs #371, #375, and #377. Until those are merged this will look like it has a lot more changes than it does. A version without all those extra changes can be found [here](https://github.com/bcaddy/cholla/pull/16/commits).


src/integrators/VL_3D_cuda.cu
```diff
@@ -158,10 +157,10 @@ void VL_Algorithm_3D_CUDA(Real *d_conserved, Real *d_grav_potential, int nx, int
   hipLaunchKernelGGL(HIP_KERNEL_NAME(Calculate_Roe_Fluxes_CUDA<reconstruction::Kind::pcm, 0>),
                      roe_pcm_launch_params.get_numBlocks(), roe_pcm_launch_params.get_threadsPerBlock(), 0, 0,
                      dev_conserved, Q_Lx, Q_Rx, F_x, nx, ny, nz, n_cells, gama, n_fields);
-  hipLaunchKernelGGL(HIP_KERNEL_NAME(Calculate_Roe_Fluxes_CUDA<reconstruction::Kind::pcm, 1>),
+  hipLaunchKernelGGL(HIP_KERNEL_NAME(Calculate_Roe_Fluxes_CUDA<reconstruction::Kind::pcm, 0>),
```
evaneschneider: Is this intentional? Aren't the directions supposed to iterate?

helenarichie (changes requested):
Except for the issue raised in Evan's comment, this looks good to me.

src/integrators/VL_3D_cuda.cu
bcaddy: That's a bug, good catch. I fixed it.

bcaddy:
I believe I've addressed everything, please let me know if I haven't.

src/integrators/VL_3D_cuda.cu
helenarichie: Looks great, thanks!

helenarichie (approved):

PR #387, opened 2024-03-26
Updating Galaxy ICs and gravity

mabruzzo:
This PR is now ready for review (modulo some changes to this description). This PR should be reviewed before #386.

# Overview
The main changes I introduced in this PR include:

1. Updating the Ideal Galaxy Initialization.
2. Updating relevant particle initialization routines
3. Updating some stuff about gravity routines (a minor tweak and some minor refactoring)

Most of the other stuff was inherited from work done by Orlando (and Alwin). These other areas are listed below.
- moving feedback routines from the supernovae directory to feedback (please don't fixate on this section of the codebase. PR #386 radically refactors/reorganizes all of this logic)
- tweaks to the feedback-analysis module
- adding extra (currently unused) logic for cooling
- adding the ability to place a temperature floor/ceiling (I refactored the logic here to some degree)

# Detailed Description: Updating the Ideal Galaxy Initialization

1. Centralizing a lot of the functionality for computing quantities related to the analytic, static gravitational potential.
    - Before this pull request, a lot of this functionality was duplicated. There were a bunch of global functions defined within `disk_ICs.cpp` and there were a handful of functions defined as methods of `DiskGalaxy`. These functions had similar names and were identical (I recall that there were minor differences in variable names, but there was no difference in logic). Some of this functionality was also replicated in other parts of the codebase.
    - In this PR, I centralized all of these functions in one place. The functionality is encapsulated by a set of `Potential` classes (e.g. there's a class for an NFW potential and a class for a Miyamoto-Nagai disk) that are internally stored by the `DiskGalaxy` class
    - The `DiskGalaxy` class be used to call this functionality in `.cpp` files.
    - The underlying `Potential` classes can be extracted from the `DiskGalaxy` class and called directly in `.cu` files (the functionality can be used on the host or the device)
2. To facilitate the usage of this functionality in both `.cpp` and `.cu` files I needed to move the implementation of the `DiskGalaxy` methods to a separate source file and the set of `Potential` classes into a separate header that is not explcitly included within `disk_galaxy.h`. This lets `.cpp` files include `disk_galaxy.h` without encountering cuda-specific extensions
3. I did a lot of work on implementing more stable initial conditions that tries to account for the initial self-gravity of the gas disk.
4. I made a lot of other structural changes within `disk_ICs.cu`.

[...]

PR #393, opened 2024-05-22
Introducing ``ParameterMap``

mabruzzo:
Adds a new ``ParameterMap`` class to assist with parameter parsing.

### Motivation

Currently, Cholla works by storing runtime parameters within a global `Parameters` struct/class and then the instance of the `Parameters` struct gets passed around to various parts of the code, which read needed values from the struct. While there is nothing wrong with our current system (it definitely gets the job done!), there are a couple of drawbacks.

For example:
- to add a new parameter, you need to introduce a new parameter to the `Parameters` struct and you need  to add parsing code to `global.cpp`. This can be somewhat undesirable since the list of parameters may be quite separated from the logic that needs it.
- it introduces a bunch of ifdef-statements within the `Parameters` struct in the cases where a parameter is only used in certain configurations. (this logic is needed to check for unused parameters)

Athena++ (and more recently, Enzo-E) adopt a cool solution where they initialize an instance of a class that stores all the arbitrary collection of specified parameters and their values (Athena++ calls it `ParameterInput`). 
- Unlike, our `Parameters` struct this object doesn't have any expectations about what the expected parameters are or should be.
- They then pass around this object to the various "code-modules" and have the modules access the parameters they need through this object.
- Introducing a new parameter to a given "code-module" is as simple as having that code-module check for the name of that new parameter. Consequently:
  - the names of parameters needed by a given module are located in a similar code-location
  - if you use disable a given module with an ifdef statement, you are also disabling the parsing of the parameters needed by that module.

This PR doesn't try to do all of that. Instead, we essentially take the first step towards that system. We introduce a class like `ParameterInput`.

### Overview

This class introduces the ``ParameterMap`` class. This class is used to store the contents of a parameter file like a "map" (or a python dictionary). It provide methods to access the values associated with a given parameter along with the expected type. It also has the ability to specify a default value.

In the future, the idea would be to pass around an instance of this map to configure various "submodules". For example, it would be easy to do this to specify runtime args for the ``feedback`` module introduced by PR #386 or the galaxy ICs in #387 or configuring the grackle module.


mabruzzo:
From the failing tests, I assume I made a logical error somewhere. My best guess is that it's related to reading in options passed on the command line? (if so that could have happened during the rebase)

I need to fix that issue. But otherwise, I think people can still review this PR (I don't think anything will change much)

src/global/global.cpp
```diff
@@ -101,137 +102,58 @@ char *Trim(char *s)
...
-    if (strcmp(param_name, optionalParam) == 0) {
-      return 1;
-    }
-  }
-  return 0;
-}
+bool Old_Style_Parse_Param(const char *name, const char *value, struct Parameters *parms);
 
-void Parse_Param(char *name, char *value, struct Parameters *parms);
+void New_Style_Init_Param_Struct_Members(ParameterMap &param, struct Parameters *parms);
```
ojwg: How about naming this `Init_Param_Struct_Members`, since `New_Style` never ages well.  `Old_Style_Parse_Param` is OK since at some point it'll be deprecated.

src/global/global.cpp
```diff
@@ -495,7 +405,54 @@ void Parse_Param(char *name, char *value, struct Parameters *parms)
...
+ *
+ *  The goal is eventually get rid of the old-style function
+ */
+void New_Style_Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
+{
+  // load the domain dimensions (abort with an error if one of these is missing)
+  parms->nx = pmap.value<int>("nx");
+  parms->ny = pmap.value<int>("ny");
+  parms->nz = pmap.value<int>("nz");
+  CHOLLA_ASSERT((parms->nx <= 0) or (parms->ny <= 0) or (parms->nz <= 0), "domain dimensions must be positive");
```
ojwg: are we asserting that at least one domain dimension must be zero or negative?  Shouldn't this be `(parms->nx > 0) and (parms->ny > 0) and (parms->nz > 0)`?  (from talking to you it seems you already caught this.  You can probably ignore this)

ojwg (approved):
2 items, but Matthew has already addressed them

PR #399, opened 2024-06-20
Updated lux make host and setup files.

brantr:
Simple updates to the host and setup files for lux.


mabruzzo (approved):
Seems like a no-brainer

PR #400, opened 2024-06-21
Allow for non standard cosmologies, and improve cosmological time integration

brantr:
We want to include non-standard cosmological models that have time-dependent dark energy equation of state. This pull request implements w0,wa parameterization of the DE EOS. It also incorporates a Runge-Kutta integrator to improve the integration of the physical timestep with scale factor, pre-computes the offset in physical time for z<\infty ICs, and writes the expansion history to a file. We also allow for non-zero radiation energy density.  I have checked the expansion history agrees to within ~0.1% with independent calculations.


brantr:
The cpp-format test fails on the following code block in `global/global.cpp`

```
// NOLINTNEXTLINE(cert-err58-cpp)
// NOLINTNEXTLINE(*)
const std::set<std::string> optionalParams = { "flag_delta","ddelta_dt","n_delta","Lz","Lx","phi","theta","delta","nzr","nxr","H0","Omega_M","Omega_L","Omega_R","Omega_K","w0","wa","Init_redshift","End_redshift", "tile_length", "n_proc_x", "n_proc_y", "n_proc_z" }; // NOLINT
```

This line already had a NOLINTNEXTLINE(), but I could not get that to work no matter how I formatted the line. I have tried various iterations of NOLINTBEGIN/NOLINTEND, NOLINT, NOLINTNEXTLINE, etc.  So if you can recommend how to get around the clang formatting issue, that would be much appreciated.

src/cosmology/cosmology.h
```diff
@@ -19,6 +19,9 @@ class Cosmology
   Real Omega_L;
   Real Omega_K;
   Real Omega_b;
+  Real Omega_R;
+  Real w0;
+  Real wa;
```
mabruzzo, line 24: One idea to  consider in the future (definitely not necessary for this PR) -- maybe it would be useful to make a new struct for representing the parameterization of the cosmology (as specified in the parameter file). 

Basically I'm describing something like:

```c++
struct CosmoSpecification {
  Real H0;
  Real Omega_M;
  Real Omega_L;
  Real Omega_K;
  Real Omega_b;
  Real w0;
  Real wa;
};
```

We could track a copy of this struct inside the ``Cosmology`` class and then we could pass this struct to the various helper functions, instead of passing long lists of parameters.

src/cosmology/cosmology_functions.cpp
```diff
@@ -20,29 +24,69 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
   chprintf("Cosmology Successfully Initialized. \n\n");
 }
 
+Real Cosmology::dtda_cosmo(Real da, Real a)
+{
+  // Return dt/da * da
```
mabruzzo: Minor formatting suggesting in case we ever want to generate doxygen documentation

```suggestion
/* Computes dt/da * da */
Real Cosmology::dtda_cosmo(Real da, Real a)
{
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -20,29 +24,69 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
+  Real a2     = a * a;
+  Real fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
+  Real a_dot  = sqrt(Omega_R / a2 + Omega_M / a + a2 * Omega_L * fac_de + Omega_K) * H0;
+  return da / a_dot;
+}
+
+Real Cosmology::Get_dt_from_da_rk(Real da, Real a)
+{
+  // Return dt/da * da
+  // But compute dt/da using a Runge-Kutta integration step
```
mabruzzo: Minor formatting suggesting in case we ever want to generate doxygen documentation

```suggestion

/* Compute dt/da * da. dt/da is computed with a Runge-Kutta integration step */
Real Cosmology::Get_dt_from_da_rk(Real da, Real a)
{
  
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -130,4 +174,64 @@ void Grid3D::Change_GAS_Frame_System(bool forward)
...
+/* local function that designates whether we are using a root-process. It gives
+ * gives a sensible result regardless of whether we are using MPI */
+static inline bool Is_Root_Proc()
+{
+  #ifdef MPI_CHOLLA
+  return procID == root;
+  #else
+  return true;
+  #endif
+}
```
mabruzzo: Since this function is already defined ``io.cpp``, I would strongly encourage you to consolidate both versions of the function. Maybe it would be easiest to move the definition to ``io.h``?

src/cosmology/cosmology_functions.cpp
```diff
@@ -130,4 +174,64 @@ void Grid3D::Change_GAS_Frame_System(bool forward)
...
+
+void Cosmology::Create_Expansion_History_File(struct Parameters *P)
+{
+  // writing the expansion history
```
mabruzzo: Minor suggestion for docstring-formatting

```suggestion
/* create the file for recording the expansion history */
void Cosmology::Create_Expansion_History_File(struct Parameters *P)
{
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -130,4 +174,64 @@ void Grid3D::Change_GAS_Frame_System(bool forward)
...
+  message += " " + std::to_string(Omega_R) + " " + std::to_string(Omega_K);
+
+  std::ofstream out_file;
+  out_file.open(file_name.c_str(), std::ios::app);
+  out_file << "# Run date: " << dt;
+  out_file << message.c_str() << std::endl;
+  out_file.close();
+}
+
+void Cosmology::Write_Expansion_History_Entry(void)
```
mabruzzo, line 218:
```suggestion
/* Write the current entry to the expansion history file */
void Cosmology::Write_Expansion_History_Entry(void)
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -20,29 +24,69 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
+Real Cosmology::Get_dt_from_da(Real da, Real a)
 {
-  Real a2    = current_a * current_a;
-  Real a_dot = sqrt(Omega_M / current_a + a2 * Omega_L + Omega_K) * H0;
-  return da / a_dot;
+  return Get_dt_from_da_rk(da, a);
+  /* Real a2     = a * a;
+  Real fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
+  Real a_dot  = sqrt(Omega_R / a2 + Omega_M / a + a2 * Omega_L * fac_de + Omega_K) * H0;
+  return da / a_dot; */
```
mabruzzo, line 81: Do we need to save this? If so, maybe it would be helpful to add an explanatory comment for why this is here.

src/cosmology/cosmology.cpp
```diff
@@ -42,8 +45,24 @@ void Cosmology::Initialize(struct Parameters *P, Grav3D &Grav, Particles3D &Part
...
+  while (a_t_sec < current_a) {
+    if (a_t_sec + da_t_sec > current_a) {
+      da_t_sec = current_a - a_t_sec;
+    }
+    dt_physical = Get_dt_from_da(da_t_sec, a_t_sec);
+    t_secs += dt_physical * time_conversion;
+    a_t_sec += da_t_sec;
+    // chprintf(" Revised a_t_sec %f da_t_sec %f t_start : %e Myr\n", a_t_sec, da_t_sec, t_secs / MYR);
+  }
+  chprintf(" Revised t_start : %f Myr\n", t_secs / MYR);
```
mabruzzo: Would you please add a comment or two to this block of code clarifying the purpose?
- specifically, could you describe the purpose of this logic? It seems like the goal is to modify `t_secs`.
   - are we just setting ``t_start`` to the age of the universe at ``current_a``? (If so, why didn't we do this before?)
   - or are we doing something else? 
- Could you address what the while loop is specifically doing?

mabruzzo (changes requested):
So this generally seems sensible to me. 

There are 2 things I would appreciate you addressing before I approve: (i) my comment inside ``Cosmology::Initialize`` and (ii) my comment about ``Is_Root_Proc``.

Once you do that, I'm happy to approve. (The other comments all provide minor suggestions -- if you're opposed to them feel free to ignore them)

mabruzzo:
> The cpp-format test fails on the following code block in `global/global.cpp`
> 
> ```
> // NOLINTNEXTLINE(cert-err58-cpp)
> // NOLINTNEXTLINE(*)
> const std::set<std::string> optionalParams = { "flag_delta","ddelta_dt","n_delta","Lz","Lx","phi","theta","delta","nzr","nxr","H0","Omega_M","Omega_L","Omega_R","Omega_K","w0","wa","Init_redshift","End_redshift", "tile_length", "n_proc_x", "n_proc_y", "n_proc_z" }; // NOLINT
> ```
> 
> This line already had a NOLINTNEXTLINE(), but I could not get that to work no matter how I formatted the line. I have tried various iterations of NOLINTBEGIN/NOLINTEND, NOLINT, NOLINTNEXTLINE, etc. So if you can recommend how to get around the clang formatting issue, that would be much appreciated.

@bcaddy -- do you have thoughts on this?

src/cosmology/cosmology_functions.cpp
brantr: I'll default to this comment format in the future.

src/cosmology/cosmology_functions.cpp
brantr: I'll default to this format in the future.

src/cosmology/cosmology_functions.cpp
brantr: I just didn't want to touch io.* in this PR. I can move the definition to io.h.  I'll update with a separate commit.

src/cosmology/cosmology_functions.cpp
brantr: I'll default to this format in the future.

src/cosmology/cosmology_functions.cpp
brantr: I'll default to this format in the future.

src/cosmology/cosmology_functions.cpp
brantr: I'll add a comment in the next commit. I'd like to save it for now, so I can revert to the Euler integration in testing if needed.

src/cosmology/cosmology.cpp
brantr: I'll add a comment in the next commit.

I am indeed just modifying t_start to the age of the universe at current_a.  This is not so important for LCDM cosmologies, but when checking the expansion history for a non-standard cosmology it is very helpful.  Another issue is that the ICs will not always start at the same redshift, so `t_secs` is not measured relative to z~\infty.

The while loop advances the integration until the scale factor corresponding to `t_secs` is the same as the initial scale factor of the ICs (or restart) file.

src/cosmology/cosmology.h
brantr: I've been thinking about how best to clean this up. It will also impact the cooling and RT portions of the code, but those need to be reconciled anyway. I will come back to this in the future.

src/cosmology/cosmology_functions.cpp
brantr: I've addressed this in a commit 5d3c10ba

src/cosmology/cosmology_functions.cpp
brantr: I've address this in commit 2b3b2a0f

src/cosmology/cosmology.cpp
brantr: I've addressed this in commit 2b3b2a0f

src/io/io.h
```diff
@@ -7,6 +7,17 @@
 #include "../global/global.h"
 #include "../grid/grid3D.h"
 
+/* Local function that designates whether we are using a root-process. It gives
+ *  * gives a sensible result regardless of whether we are using MPI */
```
mabruzzo, line 11: One formatting suggestion (and fixing a typo -- the typo is leftover from when I originally wrote the docstring)
```suggestion
 *  a sensible result regardless of whether we are using MPI */
```

mabruzzo (approved):
Looks great to me! There is 1 last formatting thing suggestion.

After we deal with the linting error (@bcaddy or @helenarichie any ideas?) we'll be good to merge. I can try to take a more careful look at the linting error tomorrow

brantr:
After a lot of effort and help from Bruno, the version 1f05b8d reproduces well the original cosmological hydro sims using the same flags. The salient issue that may need discussion is the last commit that undoes changes to PPMP_cuda.cu, where ifdef CTU had been changed to ifndef VL. The last commit undoes this change, since it was affecting the cosmological hydro integration, and these portions of the PPMP scheme now only apply to CTU again (not VL or SIMPLE). @evaneschneider do you know why this change was made?  Is it OK to keep this off whenever SIMPLE is used, or should I add an ifndef COSMOLOGY and revert the changes back to ifndef VL?

evaneschneider:
@brantr I do think this needs some discussion. Perhaps part of the problem is that I don't really understand what SIMPLE should do. As I've mentioned a couple times, I think SIMPLE is numerically unstable in more than 1D, since I thought it was just the first half of CTU, and removing the characteristic projections and time evolution of the interface states makes it numerically unstable in 1D as well, so I probably thought I was helping when I noticed that step had been eliminated. (I do remember showing some KH tests in the CAAR slack that demonstrated instabilities that went away when the characteristic projection step was turned back on.) That said, if you all have worked out that simple works without those steps, we can (and perhaps should) get rid of all of that code (and I think we should be consistent here, either we have it in all the reconstructions, or we have it in none). I think resolving this question is best done through a conversation, since it seems like maybe you all did some development work in creating simple that I'm unaware of / do not understand.

brantr:
Let's discuss offline, but then when we resolve we can post back here for prosperity.

evaneschneider:
When you get a chance @brantr can you re-run clang-format? I was going to try to investigate the linting error but right now there are a bunch of violations in a bunch of files.

evaneschneider:
@brantr Are you sure you're running the right version of clang-format? There are still a bunch of errors in a number of files, but a lot of them seem to be simple things (like not removing end spaces or not getting equalities in line) that should just be taken care of when you run the formatting, so I don't really understand why it's not working.

evaneschneider:
BTW, other than the formatting errors, everything seems okay on this branch except the new cosmology system test, which makes sense. Given that I assume this version has been verified for some forms of correctness, I'm happy to merge this and then rerun the system test and swap out the test data.

evaneschneider:
pre-commit.ci autofix

mabruzzo:
Oh wow! it looks like it precomit worked! That's extremely encouraging!

evaneschneider:
Regarding formatting, thanks to @mabruzzo's PR #404 we now have the ability to fix formatting errors on PRs, so I went ahead and applied those changes.

PR #402, opened 2024-07-01
Remove line introducing a bug in cosmological cooling unit system.

brantr:
In chemistry_functions.cpp, relative to Bruno's cosmology version, there are changes to the unit system to allow for the CHEMISTRY_GPU modules to be used without COSMOLOGY set. The changes couch scale factor scalings or little h conversions within ifdef COSMOLOGY statements.

On line 62, there is a dens_number_conv that is scaled by a^3. In Bruno's code, this undoes a a^-3 of density_units. However, in this version of the dens_number_conv is set directly by dens_CGS, not density_units, and so the additional factor of a^3 is incorrect.

I have verified that removing line 62 addresses the bug and the results show good agreement with Bruno's version* (* I have only checked with DE off to fairly high redshift, but enough to diagnose this issue).


brantr:
Noting that this change is just removing a line within an ifdef COSMOLOGY, and it won't affect any other builds.

PR #406, opened 2024-07-22
Improving thread crash logic

mabruzzo:
This introduces 3 changes to the behavior revolving around fixing thread-crashes (i.e. density or energy drops to a negative value). For the uninitiated, the default behavior in a thread-crash, when `DENSITY_FLOOR` isn't defined, is to call `Average_Cell_All_Fields`, which overrides the values in the crashed cell with the values from the neighboring cells.

The 3 changes include:

1. **A major synchronization bugfix:**
   - we moved the call to `Average_Cell_All_Fields` from the `Update_Conserved_Variables_3D` kernel to the newly created `PostUpdate_Conserved_Correct_Crashed_3D` kernel. We always launch `PostUpdate_Conserved_Correct_Crashed_3D` after `Update_Conserved_Variables_3D`
   - Previously, we could be averaging the values of neighboring cells while they were in the middle of an update. 
   - This was producing large spikes in Energy (1000 times larger than a supernova) in our simulations. This fix seemed to fix a little more than I was expecting (I still need to finish examining that).
2. A minor bugfix:
   - Previously `Average_Cell_All_Fields` always looked for neighbors, regardless of where the central cell was located on a grid. So if you were at the edge of the active zone, you might include the values of cells in the ghost-zone. This was problematic when dealing with a crashed-cell at the hydro update when the ghost zones had not been updated yet (it's unclear if this commonly happened).
   - Care was taken to maintain existing behavior in `Average_Slow_Cells`, which could apply `Average_Cell_All_Fields` to a slow-cell at identified at any location on the grid (in the ghost zone or active zone).[^1] My change also happens
3. A minor improvement:
   - Previously the `Average_Cell_All_Fields` would ignore any neighboring cells with a non-positive density or pressure. 
   - This PR extends the behavior to also ignore any cells that would be considered a "slow" cell and will need to be separately averaged. (When the `AVERAGE_SLOW_CELLS` macro isn't defined, no cell is considered a "slow" cell)

The implementation of the third change got a little hairy. In principle I think it's a good design choice (especially for reducing ifdef statements). But, with the benefit of hindsight, maybe I should have done something simpler. Definitely open to feedback on this point!

[^1]: In this case we continue for a cell on the edge of the active zone, we will include cells in the ghost zone in the average. As I write this out, I'm not totally sure the ghost zones are refreshed between the invocation of the Hydro-Solver and the call to `Average_Slow_Cells`. **That is something to followup on!**

### DIFF
--- a/src/grid/grid3D.cpp
+++ b/src/grid/grid3D.cpp
@@ -9,6 +9,7 @@
 #include "../global/global.h"
 #include "../grid/grid3D.h"
 #include "../grid/grid_enum.h"    // provides grid_enum
+#include "../hydro/average_cells.h"  // provides Average_Slow_Cells and SlowCellConditionChecker
 #include "../hydro/hydro_cuda.h"  // provides Calc_dt_GPU
 #include "../integrators/VL_1D_cuda.h"
 #include "../integrators/VL_2D_cuda.h"
@@ -152,7 +153,9 @@ void Grid3D::Initialize(struct Parameters *P)
 
 #ifdef AVERAGE_SLOW_CELLS
   H.min_dt_slow = 1e-100;  // Initialize the minumum dt to a tiny number
-#endif                     // AVERAGE_SLOW_CELLS
+#else
+  H.min_dt_slow = -1.0;
+#endif  // AVERAGE_SLOW_CELLS
 
 #ifndef MPI_CHOLLA
 
@@ -578,7 +581,8 @@ Real Grid3D::Update_Hydro_Grid()
   ny_off = ny_local_start;
   nz_off = nz_local_start;
   #endif
-  Average_Slow_Cells(C.device, H.nx, H.ny, H.nz, H.n_ghost, H.n_fields, H.dx, H.dy, H.dz, gama, max_dti_slow, H.xbound,
+  Average_Slow_Cells(C.device, H.nx, H.ny, H.nz, H.n_ghost, H.n_fields, gama,
+                     SlowCellConditionChecker(max_dti_slow,H.dx, H.dy, H.dz), H.xbound,
                      H.ybound, H.zbound, nx_off, ny_off, nz_off);
 #endif  // AVERAGE_SLOW_CELLS
 
--- a/src/grid/grid3D.cpp
+++ b/src/grid/grid3D.cpp
@@ -454,7 +454,7 @@ void Grid3D::Execute_Hydro_Integrator(void)
 #ifdef SIMPLE
     Simple_Algorithm_3D_CUDA(C.device, C.d_Grav_potential, H.nx, H.ny, H.nz, x_off, y_off, z_off, H.n_ghost, H.dx, H.dy,
                              H.dz, H.xbound, H.ybound, H.zbound, H.dt, H.n_fields, H.custom_grav, H.density_floor,
-                             C.Grav_potential, SlowCellConditionChecker(1.0 /H.min_dt_slow,H.dx, H.dy, H.dz));
+                             C.Grav_potential, SlowCellConditionChecker(1.0 / H.min_dt_slow, H.dx, H.dy, H.dz));
 #endif  // SIMPLE
   } else {
     chprintf("Error: Grid dimensions nx: %d  ny: %d  nz: %d  not supported.\n", H.nx, H.ny, H.nz);
--- a/src/grid/grid3D.cpp
+++ b/src/grid/grid3D.cpp
@@ -454,7 +454,7 @@ void Grid3D::Execute_Hydro_Integrator(void)
 #ifdef SIMPLE
     Simple_Algorithm_3D_CUDA(C.device, C.d_Grav_potential, H.nx, H.ny, H.nz, x_off, y_off, z_off, H.n_ghost, H.dx, H.dy,
                              H.dz, H.xbound, H.ybound, H.zbound, H.dt, H.n_fields, H.custom_grav, H.density_floor,
-                             C.Grav_potential);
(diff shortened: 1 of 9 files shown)


mabruzzo:
pre-commit.ci autofix

src/integrators/simple_3D_cuda.cu
```diff
@@ -154,6 +155,11 @@ void Simple_Algorithm_3D_CUDA(Real *d_conserved, Real *d_grav_potential, int nx,
                      zbound, dt, gama, n_fields, custom_grav, density_floor, dev_grav_potential);
   GPU_Error_Check();
 
+  // Step 3b: Address any crashed threads
+  hipLaunchKernelGGL(PostUpdate_Conserved_Correct_Crashed_3D, dim1dGrid, dim1dBlock, 0, 0, dev_conserved, nx, ny, nz,
+                     x_off, y_off, z_off, n_ghost, gama, n_fields, slow_check);
+  GPU_Error_Check();
```
helenarichie, line 161: This is very minor, but there's a `GPU_Error_Check` following this kernel launch in the Simple integrator, but not in Van Leer, so it might be good to either remove this one or add one to VL so that the behavior is consistent between the two.

src/hydro/average_cells.h
```diff
@@ -0,0 +1,54 @@
...
+#endif
+  {
+  }
+
+  /*! \brief Returns whether the cell meets the condition for being considered a slow cell that must
+   *  be averaged. */
+  template <bool verbose = false>
+  __device__ bool is_slow(Real E, Real d, Real d_inv, Real vx, Real vy, Real vz, Real gamma) const
+  {
+    return this->max_dti_if_slow(E, d, d_inv, vx, vy, vz, gamma) >= 0.0;
```
helenarichie, line 34: Another minor question/comment, but is it possible for `max_dti_if_slow` to ever be zero here? If not, just for the sake of clarity, it might be good to change the condition to `max_dti_if_slow > 0` instead of `max_dti_if_slow >= 0`.

I know that `hydroInverseCrossingTime` can return zero, but I still don't think it's possible for `(max_dti > max_dti_slow) ? max_dti : -1.0;` to evaluate to zero unless someone sets `max_dti_slow` to a negative number. Again, very minor, but I found it a little confusing.

In line with my comment about setting `max_dti_if_slow`, requiring that it's positive would make it so that `(max_dti > max_dti_slow) ? max_dti : -1.0;` never evaluated to zero and clarify this ambiguity.

src/grid/grid3D.cpp
```diff
@@ -152,7 +153,9 @@ void Grid3D::Initialize(struct Parameters *P)
 
 #ifdef AVERAGE_SLOW_CELLS
   H.min_dt_slow = 1e-100;  // Initialize the minumum dt to a tiny number
```
helenarichie, line 155: I definitely don't think this needs to happen with this PR, but Evan and I have talked about how it would be good to set `min_dt_slow` from the input parameter file instead of at compile time. Additionally, as I mentioned below, I think it would be good to require that it's a positive number. This is definitely beyond the scope of this PR, so feel free to ignore-just wanted to make a note of it while we're overhauling cell averaging!

src/hydro/hydro_cuda.cu
```diff
@@ -711,25 +738,26 @@ __global__ void Average_Slow_Cells_3D(Real *dev_conserved, int nx, int ny, int n
...
     E     = dev_conserved[4 * n_cells + id];
 
-    // Compute the maximum inverse crossing time in the cell
-    max_dti = hydroInverseCrossingTime(E, d, d_inv, vx, vy, vz, dx, dy, dz, gamma);
+    // retrieve the max inverse crossing time in the cell if the cell meets the threshold for being a slow-cell.
+    // (if the cell doesn't meet the threshold, a negative value is returned instead)
+    max_dti = slow_check.max_dti_if_slow(E, d, d_inv, vx, vy, vz, gamma);
 
-    if (max_dti > max_dti_slow) {
+    if (max_dti >= 0) {
```
helenarichie, line 745: Again, I think `max_dti > 0` would be clearer here.

src/hydro/average_cells.h
helenarichie, line 1: Style-related comment since this is a solitary bit of new code-it would be good for the naming of the physics variables in here to follow the style guide.

helenarichie (changes requested):
I think this looks great! I have a few minor comments, but other than those I think it's ready to go.

src/integrators/simple_3D_cuda.cu
mabruzzo: Good catch! I explicitly thought I fixed that (but I must have seen it and not acted on it)

src/grid/grid3D.cpp
mabruzzo: Absolutely! Especially since most of the logic that actually sets `min_dt_slow` is currently buried in the gravity functions.

src/integrators/simple_3D_cuda.cu
mabruzzo: Turns out that this was an issue. This was fixed in #417

src/hydro/average_cells.h
mabruzzo: done in #417

src/hydro/hydro_cuda.cu
mabruzzo: done in #417

mabruzzo:
The plan is to merge in #417 instead of this PR (since #417 fixes a bug introduced here)

PR #411, opened 2024-07-26
`concat.py` now ensures outdir exists

mabruzzo:
Previously, `concat.py` would fail if the output directory didn't already exist. Now it will try to create the output directory. This intentionally refuses to recursively create directories.
- In my experience, refusing to recursively create directories is a good way to catch typos (while providing convenince)
- If recursive creation of directories is a desired behavior, we can easily add it.


evaneschneider (approved):
Looks good!

PR #412, opened 2024-07-26
Make it possible to run Cholla with particles but no gravity

mabruzzo:
### Overview

This PR makes it possible to run Cholla in a configuration where hydro and particles are enabled while gravity is disabled. This is useful for testing feedback.

A number of small, simple changes were required. These are mostly self-explanatory.

The most important change relates to `Particles3D::initialize`:
- previously `Particles3D::initialize` expected `Grav3D&` as an argument
- since the function only used `Grav3D&` to access spatial properties, it was relatively easy to change
- I created the new `SpatialDomainProps` class to be used to convey spatial properties of the grid. It provides a method to create a fresh instance by reading the spatial properties from the main grid object. We now pass this object in place of `Grav3D` to the `Particles3D::initialize`
- While I was making these changes, I also modified `Grav3D::Initialize` to accept `SpatialDomainProps` as an argument rather than passing all of the details as individual arguments.

### Future Ideas

Ultimately, in the future, I think it would be great to make the `SpatialDomainProps` class take on more importance:
- I think it would be great to directly store instances of `SpatialDomainProps` directly inside of `Grid3D`, `Particle3D`, and `Grav3D` rather than the current solution where we have 3 different approaches for storing the same data.
- I also think it would be great to create `SpatialDomainProps::Get_Position` (and have `Grid3D::Get_Position` forward onto `SpatialDomainProps::Get_Position`). An added bonus of doing this is we could consider supporting `SpatialDomainProps::Get_Position` on GPUs (frankly, we might want to do a little more refactoring in that case so we don't need to copy the entirety of `SpatialDomainProps` on GPUs, but that is very doable)
- Overall, I think this would be a great step towards reducing the complexity of `Grid3D`, `Grav3D`, `Particle3D`, and reducing code duplication


src/gravity/grav3D.cpp
```diff
@@ -16,39 +16,38 @@
 
 Grav3D::Grav3D(void) {}
 
-void Grav3D::Initialize(Real x_min, Real y_min, Real z_min, Real x_max, Real y_max, Real z_max, Real Lx, Real Ly,
-                        Real Lz, int nx, int ny, int nz, int nx_real, int ny_real, int nz_real, Real dx_real,
-                        Real dy_real, Real dz_real, int n_ghost_pot_offset, struct Parameters *P)
+void Grav3D::Initialize(const SpatialDomainProps& spatial_props, Real Lx, Real Ly, Real Lz,
```
ojwg: Very minor, but since you change `struct Parameters *P` to `Parameters *P` in the signature, do you want to do the same for other functions in this commit?

src/particles/particles_3D.cpp
```diff
@@ -176,8 +182,16 @@ void Particles3D::Initialize(struct Parameters *P, Grav3D &Grav, Real xbound, Re
     #endif
 
   G.size_blocks_array = 1024 * 128;
-  G.n_cells_potential = (G.nx_local + 2 * N_GHOST_POTENTIAL) * (G.ny_local + 2 * N_GHOST_POTENTIAL) *
-                        (G.nz_local + 2 * N_GHOST_POTENTIAL);
+  {
```
ojwg, line 185: Why is there a code block here?  Is it to limit the scope of `int n_ghost_pot`?  I'm only curious, I'm not proposing any changes.

ojwg (approved):
Changes look good.  Would be great to refactor the `SpatialDomainProps::From_Grid3D` to not take the Parameters struct (and maybe introduce a `SpatialDomainProps::From_Parameters` method.  But that's future technical debt.

src/particles/particles_3D.cpp
mabruzzo: Oh, it's because this is a fairly large function with a bunch of local variables, and it is already difficult to reason about all of the local variables. I was just trying to localize `n_ghost_pot` to a single area

src/gravity/grav3D.cpp
mabruzzo: done!

PR #415, opened 2024-08-05
Make the cooling-choice a runtime parameter

mabruzzo:
This PR builds on #414. (That needs to be merged first)

## Overview

At present, there are presently 4 different methods for chemistry or heating & cooling:
> (i) chemistry-gpu, (ii) grackle, (iii) tabulated-cloudy, (iv) piecewise-cie[^1]

At present, this choice is entirely controlled at compile time. The changes in this PR make it possible to specify the tabulated-cooling or piecewise-cie at runtime (the ground-work is laid for specifying both other choices at runtime, but they still require macros).

I also made it possible to switch photo-electric heating on and off (when using tabulated-cooling *only*).

## Description

I had to do a fair bit of refactoring.

The basic control-flow is as follows:

- Before starting the simulation, the `main` function passes in the `ParameterMap` (it holds all parsed parameters into) `configure_chemistry_callback`. 
  - if the `chemistry.kind` parameter is `none` this will return `none` (unless the macros for `chemistry-gpu` or `grackle` are defined -- in that case the program aborts)
  - if the `chemistry.kind` parameter is `tabulated-cloudy` or `piecewise-cie` this returns a callback to be used for performing heating/cooling
  - if `chemistry.kind` isn't specified, the default value is picked based on the macros (this is done to avoid breaking tests)
  - If the macros are defined for `chemistry-gpu` or `grackle` then this returns nothing (this is a short-term stopgap solution)
- The callback is an object of type `std::function<void(Grid3D&)>`, which is basically a C++ function-pointer that can wrap functor-objects in addition to regular functions (this is important for retaining the state associated with any functor-objects and properly calling destructors).
-  The `main` function then passes in the callback into `Grid.Update_Hydro_Grid`. If the callback is not empty, it will always be called.

I had to do some refactoring of **src/cooling/cooling_cuda.cu** in order to make this select-able at runtime.

## Configuration parameters

The new parameters are all in the `[cooling]` parameter table. (Parameter-tables are introduced in #414 and we leverage functionality to raise errors when there are unused parameters in the cooling group)

The new parameters include:
- `cooling.kind`
   - allowed values: `none`, `chemistry-gpu`, `grackle`, `tabulated-cloudy`, `piecewise-cie`
   - default: defaults to `none` when no chemistry/heating&cooling parameters are given
   - description: when `none`, there is no chemistry/heating/cooling
- `cooling.data_file`
  - description: this can optionally be used to specify a custom path to the data file when using `tabulated-cloudy`. It is an error to specify this for any other `cooling.kind`
- `chemistry.photoelectric_heating`
  - type: boolean
  - default: `false`
  - description: this can optionally be used to introduce photoelectric-heating when using `cooling.kind` == `tabulated-cloudy`. It is an error to specify this for any other `cooling.kind`
- `chemistry.photoelectric_n_av_cgs`
  - type: float
  - default: `100.0`
  - description: when `cooling.kind` == `tabulated-cloudy` and  `chemistry.photoelectric_heating` == `true`, this can be used to specify the average number-density in the domain in cgs units (which is used to compute impact of photoelectric heating). It is an error to specify when `chemistry.photoelectric_heating` == `false`

[^1]: Let me know if you would prefer different names


src/global/global.h
```diff
@@ -50,15 +50,6 @@ typedef double Real;
...
 
-// Conserved Floor Values
-#if (defined(FEEDBACK) || defined(STAR_FORMATION)) && defined(CLOUDY_COOL)
-  #define TEMP_FLOOR 1e1    // 10K for cloudy cooling
-  #define DENS_FLOOR 14.83  // 1e-6 cm^-3
-#else
-  #define TEMP_FLOOR 1e-3
-  #define DENS_FLOOR 1e-5  // in code units
-#endif
-
```
evaneschneider, line 61: Are these floors defined elsewhere now?

evaneschneider (approved):

src/global/global.h
mabruzzo: They aren't used anymore.

I'm pretty sure they haven't been used ever since @helenarichie made the choices of floors into runtime parameters

PR #418, opened 2024-10-26
Refactor dual energy

mabruzzo:
This PR should be reviewed after #417 is merged

------

This PR essentially moves around the contents for some dual-energy routines so that we only need to maintain a single version of these function rather than 3 independent functions. Motivation for doing this is provided at the top of the new dual_energy.h file


evaneschneider:
Can someone remind me of the status of this PR?

src/hydro/dual_energy.h
```diff
@@ -0,0 +1,277 @@
...
+    int ipo         = min(xid + 1, nx - n_ghost - 1);
+    neighbor_ids[1] = ipo;
+
+    is_real_cell = (xid > n_ghost - 1 && xid < nx - n_ghost);
+
+  } else if constexpr (NDim == 2) {
+    n_cells = grid_shape.x() * grid_shape.y();
+
+    // we maintain the use of blockId for backwards compatability, but I'm not sure it does anything
+    // (i.e. can we just replace blockId with blockIdx.x as in the other 2 branches)?
```
evaneschneider: At one point in the distant past, we needed to use 2D grids to run really large 2D problems, because of CUDA limits on the 1D grid dimensions. I believe this hasn't been an issue for a long time, but you can see it reflected in the name 'dim2dgrid' for the grid dimensions in e.g. the VL_2D_cuda.cu functions, versus 'dim1dgrid' for the 3D version, even though then we set up a a 1D grid. In principle as long as we continue to only use 1D grids, replacing this with blockIdx.x would be fine, since blockIdx.y will always be 0 for a 1D grid, but that's why it's like this (and the distinction lives on in the other 2D versions of hydro functions).

src/hydro/dual_energy.h
```diff
@@ -0,0 +1,277 @@
+/*! \file hydro_cuda.h
```
evaneschneider: should be "dual_energy.h"

src/hydro/hydro_cuda.h
```diff
@@ -107,11 +101,16 @@ __global__ void Partial_Update_Advected_Internal_Energy_3D(Real *dev_conserved,
...
-
-__global__ void Select_Internal_Energy_2D(Real *dev_conserved, int nx, int ny, int n_ghost, int n_fields);
-
-__global__ void Select_Internal_Energy_3D(Real *dev_conserved, int nx, int ny, int nz, int n_ghost, int n_fields);
+/*! Returns whether a cell has crashed
+ *
+ *  \note
+ *  It probably won't come up, but it's unclear why we don't consider a density of 0 or
+ *  energy of 0 to be crashed... (I'm just keeping the logic consistent with what it used to be)
+ */
```
evaneschneider, line 105: I'm fine with adding = to the energy criterion.

evaneschneider:
This looks good to me, once the couple small things are addressed.

mabruzzo:
pre-commit.ci autofix

PR #419, opened 2024-10-28
Convert string-param parsing

mabruzzo:
Description
-----------

This patch relocates all of the string parameter parsing from ``Old_Style_Parse_Param`` to ``Init_Param_Struct_Members``. The relocated logic was converted to use the new-style of parameter parsing.
- While I was doing this, I relocated (& converted) some additional parameter-parsing logic. Essentially I did this when a string parameter was part of a small group of related parameter-logic and I wanted to avoid breaking up this group.
- Specifically, the non-string logic that was relocated includes Cosmology parameters and ANALYSIS parameters (more on this below)

Motivation
----------

The primary motivation is so that we can make the shift to syntactic typing of parameters. I've explained what this is other PRs and this parsing logic conversion needs to be made sooner or later, so I won't go into much detail.

The main reason for this PR is that I think we should make the shift to syntactic typing sooner rather than later. 
- Essentially, this shift just means that string-parameter values will need to be quoted.
- I think this change will need to be gradually phased in (since older parameter files will need to be converted).
- It would be useful to start this process sooner rather than later so that new parameter files can use syntactic typing (the longer we wait, the more conversion work that will be needed later).

Once we fully adopt syntactic typing Cholla parameter files will be fully consistent with the TOML parameter file format (it will be a strict subset). This means we could eventually replace our custom parsing logic with logic from an existing toml parsing library (there are some nice header-only libraries that we could easily drop into cholla). Using a library introduces a number of benefits (e.g. we could support parameters that specify a list of values, which is **extremely useful**). There are also some other benefits, too.

Relocated Cosmology and Analysis Parameters
-------------------------------------------

I took a lot of care to ensure that the parsing logic remains consistent.
- Historically, cosmological parameters had no default values (and minimal value sanitization). Essentially there was not error checking. Now, cosmological simulations automatically raise errors if a user forgets to specify any cosmological parameter. The lone exception is the `wa` parameter which now defaults to a value of 0 if omitted (i.e. consistent with LCDM cosmology)
- The analysis logic previously didn't have any default values. To avoid breaking Cholla with existing parameter-files, I made the parsing logic default to dummy values in most cases (so effectively, parsing of these parameters is as permissive as ever). The exception is the `skewersdir` parameter. Cholla will abort with an error if the user forgets this parameter in a simulation compiled with `-DOUTPUT_SKEWERS`


ojwg:
[pre-commit.ci](http://pre-commit.ci/) autofix

ojwg:
pre-commit.ci autofix

src/global/global.cpp
```diff
@@ -452,6 +409,65 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+  parms->Omega_L       = pmap.value<double>("Omega_L");
+  parms->Omega_b       = pmap.value<double>("Omega_b");
+  parms->Omega_R       = pmap.value<double>("Omega_R");
+  parms->w0            = pmap.value<double>("w0");
+  parms->wa            = pmap.value_or("wa", 0.0);
+#endif  // COSMOLOGY
+
+#if defined(CHEMISTRY_GPU) || defined(COOLING_GRACKLE)
+  Load_String_Param_Into_Char_Buffer(pmap, "scale_outputs_file", parms->scale_outputs_file, "");
+#endif
```
ojwg, line 457: Shouldn't `scale_outputs_file` be `UVB_rates_file` here?

ojwg (changes requested):
potential error with  UVB_rates_file handling

src/global/global.cpp
mabruzzo: thanks for catching that!! I fixed it

mabruzzo:
Well I clearly broke something. I need to circle back to this

PR #420, opened 2024-11-22
Make ti-cool selectable at runtime

mabruzzo:
This PR builds on #415 

--------

Essentially, this makes the TI cooling curve select-able at run-time. To select TI-cool, use should pass `piecewise-ti` to the `chemistry.kind` parameter (which was introduced in #415)

This PR involved some incidental refactoring:
- I replaced the `TI_cool` function with the `CoolRecipeTI` class. The functionality of `TI_cool` is now implemented by the `CoolRecipeTI::rate_cool` member-function. This was done to match how other cooling recipes are implemented
- I slightly refactored how we implement cooling for T>1e4 K for CIE cooling and TI cooling. Previously, but regions of code used the same piecewise analytic formula, but duplicated the implementation of it. Now we just implement it once.
- I refactored how we implement photoelectric heating gets implemented.
  - previously, photoelectric heating was implemented by a function `Photoelectric_Heating` and cloudy-cooling separately tracked the `n_av_cgs` parameter. If we made the `CoolRecipeTI` class mirror the architecture of the cloudy recipe, the class would also need to track `n_av_cgs`.
  - This is a little undesirable because `n_av_cgs` is an implementation detail of a particular photoelectric heating model. Robustly documenting the code's limitations would involve repeating some warning-comments. Plus it would be a little messy to replace the photoelectric heating model. 
  - Now, we encapsulate the photoelectric heating term in a `PhotoelectricHeatingModel` callable.
  - (in the future, it might be better to not store photoelectric heating separately from the rest of the cooling recipe)

While I was here, I slightly streamlined the implementation of `CoolingUpdateExecutor` (I removed an unnecessary indirection that was only in place to simplify the number of changes in #415). I also added an overview of how the code is supposed to work to the top of the file.

> [!IMPORTANT]  
> It's worth noting that the lower Temperature sections of TI-cool cooling function comes from papers where they assumed $\mu=1.25$ and the higher Temperature is for temperatures where we can safely assume $\mu=0.6$.
> - Currently, we assume $\mu=0.6$
> - We could conceivably do some kind of special renormalization to account for changes in $\mu$, but I'm a little hesitant to go down that route


evaneschneider (commented):
Generally speaking, I think this looks good, but I'm a little confused about why the photoelectric heating piece can't be implemented as a separate recipe, rather than included in the relevant cooling recipes?

evaneschneider (approved):

PR #423, opened 2025-02-07
Parameter tables testing

mabruzzo:
this is just intended to run Jenkins

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -29,7 +29,7 @@ pipeline
                     axis
                     {
                         name 'CHOLLA_MAKE_TYPE'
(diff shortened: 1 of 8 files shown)


mabruzzo:
My current speculation is that the functions within `gridReduceMax` always assumes that you are launching the reduction where the number of threads per block is an integer multiple of the warpSize and I don't think that's the case in this test case. Consequently, there is a bunch of garbage data that we are performing the reduction upon. In my local test-runs, that garbage uninitialized data consists of really small values, but I would guess that on the CI (in this particular config) it's something else... Again this is all currently a guess

PR #424, opened 2025-02-14
Fix the flaky failing test!

mabruzzo:
Explicitly initialize the output buffer in `TEST(tHYDROCalcDt3D, CorrectInputExpectCorrectOutput)`. This was already being done in the other places where `Calc_dt_[123]D` gets called.


evaneschneider (approved):
Great catch!

PR #425, opened 2025-02-14
added h5AttrsFuncs python file

mageejake:
Added h5AttrsFuncs.py to python files in dev branch

File contains 2 functions to read in and modify attributes of .h5 files


evaneschneider (changes requested):
Can you put it in the 'python_scripts' directory?

mabruzzo:
It might also be cool if you added a `if __name__ == __main__:` section that supported the use of the script as a command line tool. But, that may be beyond the scope of this PR

PR #430, opened 2025-05-21
Introduce the WriterManager type

mabruzzo:
> [!NOTE]
> The benefits of this PR won't be apparent simply by looking at the source code. This is largely foundational for a few other PRs I had planned. But, I tried to explain the benefits down below. With that said, if you're not sold on it, you could just wait for one of the followup PRs for more concrete evidence on how this helps

While this patch doesn't do a whole lot right now, it is important for achieving 2 goals:
1. A followup PR will make WriterManager directly read from `ParameterMap` in order to transition away from storing most io-parameters inside of the global Parameters struct. Some examples of how this will work:
   - we will be able to directly read "n_hydro", "n_particle", "n_slice", ... directly from `ParameterMap` within `WriterManager`'s constructor.
   - we might create a "functor" for writing 32-bit outputs. Pseudo-code to sketch out what I'm thinking of (it looks like python just to make it easier for people to understand)
      ```python
      class F32Writer:
          def __init__(self, parameter_map):
              self.density = parameter_map.read("out_float32_density")
              self.Energy = parameter_map.read("out_float32_Energy")
              # ... (repeat for momenta, bfields internal energy, etc.)

          def __class__(self, ...):
              # call a slightly modified version of Output_Float32 that
      ```
     A functor like this could easily be tracked within the `WriterManager` without any modifications. It would let us remove all all of the `out_float32_<field>` members from the definition of the `Parameters` struct
   - with a little work, we could also define for a functor for making Rotated Projections (and move all of its parameters as well)
   - It will also make introducing new configurable output writers (I know Evan was talking about "thin projections" at one point)

2. Once we are reading directly from the `ParameterMap`, a subsequent PR can easily use this framework for turning on and off various derived output types (slices, projections, rotated-projections) and we can get rid of the associated macros


evaneschneider (approved):
I think this looks great!

PR #433, opened 2025-05-28
Writer manager tweaks

mabruzzo:
A simple followup to #430. Basically, this just applies some of the changes outlined in one of the future goals of PR #430. 

We now directly read `"n_hydro"`, `"n_particle"`, `"n_slice"`, ... directly from `ParameterMap` within `WriterManager`'s constructor.


src/io/WriterManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <string>
+#include <vector>
+
+#include "../gravity/grav3D.h"
+#include "../io/ParameterMap.h"  // define ParameterMap
+#include "../io/io.h"
+
+io::WriterManager::WriterManager(const Parameters& P, ParameterMap& pmap) : fname_template_(P)
+{
+  // in the future, the goal is to read directly from ParameterMap (so we can stop storing
```
evaneschneider, line 18: should this comment be adjusted to reflect the fact that we are now reading from the parameter map?

src/io/WriterManager.cpp
```diff
@@ -0,0 +1,64 @@
...
+  };
+  packs_.push_back(io::detail::WriterPack{"particle", pmap.value_or("n_particle", 1), write_particle});
+
+#endif
+
+#if defined(GRAVITY) && defined(HDF5)
+  auto write_gravity = [](Grid3D& G, Parameters P, int nfile, const FnameTemplate& fname_template) {
+    G.Grav.Write_Restart_HDF5(&P, nfile, fname_template);
+  };
+  int n_gravity = 1;  // <- this is the historical choice
```
evaneschneider, line 60: Does this mean there is not a way to do the gravity write-out on longer intervals than the outstep? If so, perhaps we can change it to behave like the others (default to 1, but allow a different multiple to be read in the parameter struct?) Or is there some other macro or parameter that controls the frequency of writing of the potential for restarts?

src/io/WriterManager.cpp
mabruzzo: Good catch

evaneschneider (approved):
Looks good to me!

PR #434, opened 2025-06-23
Introduce the `cholla_utils` python package

mabruzzo:
This introduces functionality to the repository to make it easier to load snapshot datasets, regardless of whether
- they were concatenated with old scripts (and never repacked)
- they were concatenated with newer scripts
- the snapshot was never repacked (i.e. it is distributed across multiple hdf5 files).

To make it easier to use this function, I have introduced the functionality in the form of a python module. I have added instructions for installing the python module and basic usage instructions. A little more work is needed before this is completely done (I would like to add some slightly more rigorous tests to make sure that we continue to maintain backwards compatibility with files that were concatenated by older versions of the concatenation scripts).


python/README.md
```diff
@@ -0,0 +1,75 @@
...
+1. If you want to query the fields that are saved within the dataset, you can invoke the following snippet:
+
+   ```python
+   import cholla_utils
+
+   my_fields = cholla_utils.get_native_fields("path/to/snap.h5.0")
+   print(my_fields)  # prints out the names of all saved fields
+   ```
+
+2. Loading a single field: This is easy. Suppose the snapshot is saved at If that snapshot has a field called `"density"`, you can simply invoke:
```
evaneschneider: typo: missing path

evaneschneider (approved):
This looks good to me. Let me know when it's ready to merge.

PR #435, opened 2025-07-09
Fix the default build files on Frontier

mabruzzo:


evaneschneider:
Are these are the build file parameters from the scaling tests a few weeks ago? (So they are good as of 7/2025?)

mabruzzo:
Yes! I was using them in the context of our galaxy test problem.

For some background:
- the build-file that currently live in the repository (that I am replacing) does some weird stuff involving the `--gcc-toolchain` flag.
- This was producing compiler errors in this configuration. I had traced back the origin of these flags through the commit history (and a slack discussions), and found out that Trey had come up with that set of flags to deal with some issues that Brant encountered with the `std::filesystem` library (which is used for IO).
- I went out of my way to confirm that by updating the version of rocm (and we stopped pinning the  new configuration flags indeed compile with `std::filesystem`

Essentially, all I did was revert the flags to the older (simpler) version and started using newer versions of dependencies. I suspect that the problem stemmed from the versions of the CPE[^1] package or rocm that had been current at the time.

[^1]: The Cray Programming Environment meta package

evaneschneider:
Got it, thanks!

PR #438, opened 2025-08-04
Bump container image

mabruzzo:
This is a reissue of #437 (I literally just moved the branch to the Cholla Repository). I think this was the source of all my problems!

-------------

I've done a lot of reading in the past week (before I ultimately just started trying things out).

Initially, I tried to start with the regular ROCm image and then manually use sudo apt-get in the workflow to actually install the stuff we need... I've concluded that won't work... We need to install all build dependencies inside the container.

So then I moved onto actually building the container with the github action. Ideally we want to be able to make some tweaks to the Dockerfile and let GitHub make all of the updates for us. I spent a bunch of time tinkering around within https://github.com/cholla-hydro/cholla/pull/386, before I moved onto experimenting elsewhere...


mabruzzo:
@evaneschneider, this is ready for review.

Please take a look at the **.github/workflows/README.md** file that I added and let me know if you find any part of it to be confusing (or if you think any part of it needs clarification). The goal was to provide detailed enough instructions to make it possible for anyone to bump the docker image (and provide basic context in case somebody eventually needs to make a more significant change)

Aside: There is a small chance that I'll need to open a 2nd PR that adds the `build_image.yml` to the main branch, if the button for manually running the workflow doesn't show up after we merge this PR.[^1]

[^1]: I'm a little hesitant to directly merge this PR into main because the README talks about workflows that I previously modified (and the modifications are not present in the main branch)

evaneschneider (approved):
All looks good to me!

PR #440, opened 2025-08-18
This PR introduces some extra pre-commit sanity checks

mabruzzo:
I've added most of these checks to other projects that I work on. These are mostly just sanity checks.

These checks include:

- a check that all non-binary files marked as executable (via ``chmod``) have a ["shebang"](builds/check.sh) on the file's first line.

  - by adding this, I actually was able to catch a few minor mistakes

- a check that identifies sets of files that would produce errors on a case-insensitive file system (this exact problem caused some minor headaches for me at one point with Enzo-E)

- check to confirms that .toml files (like pyproject.toml) and .yml/.yaml files (commonly used for configuring ci) have valid syntax


evaneschneider:
What's going on with the empty text files in the 'feedback' folder? I think that folder doesn't exist in the dev branch yet.

mabruzzo:
The text files already exist in the dev branch alongside the initial version of the feedback module that Orlando had written. (I think this got merged in with either the gravity or the initial conditions PR since it was too difficult to disentangle things).

The files are not in fact empty; I just confirmed by checking them in branch that I am proposing to merge. I think the GitHub file-diff viewer is just indicating (in a confusing manner) that none of the contents of the file have changed. The only thing that is is changing is the file permissions (I think that they previously had executable permissions, which makes no sense for text files)

evaneschneider (approved):

PR #442, opened 2025-08-18
Apply the python code formatter to a subset of python scripts

mabruzzo:
This commit primarily applies the python code formatter (using the same settings that we use in our ``cholla_utils`` python package) to a subset of the files in ``python_scripts``. In a few isolated cases I tweaked the formatting of a docstring by hand.

**No python logic was changed.**

I chose to format these particular files because I plan to move them finto the cholla_utils package or have them live as scripts that are actively maintained in close association with the cholla_utils package. (This happens in PR #443, which should be reviewed after this PR). It will be easier to do this if all these files can be formatted/linted with the same tools as the rest of the ``cholla_utils`` package.

**Note:** the formatting style changed a lot in a small subset of these files. In those files, the new style is the de-facto standard for open source python code (so I think it improves readability).

**Aside:** I don't plan to reformat any other files from ``python_scripts``. The plan is to move these to a separate directory in a subsequent PR, and retain them for posterity. But, there's little need to maintain them (I don't think anybody actually uses them).


evaneschneider:
Looks good!

evaneschneider (approved):

PR #448, opened 2025-09-03
Transfering some documentation from the Wiki

helenarichie:
Transferred the docs for the dust, MHD, and dual energy physics modules (with some small tweaks to convert Markdown syntax to MyST Markdown syntax), and added the command for viewing the local version of the documentation site in Chrome on macOS.


mabruzzo (approved):
Looks great to me!

PR #450, opened 2025-09-15
Add the `par` sphinx-extension

mabruzzo:
This PR introduces the `par` sphinx-extension to help us with documenting our runtime parameters.

This is a sphinx extension that I originally developed for Cello/Enzo-E. I have made some small generalizations so that it will work well for Cholla. (Aside: I'm actually thinking about making it a standalone python module since that wouldn't take much work at this point).

-----

I want to make the documentation of the parameters a little more compact. I have a few ideas on how to do that, but I don't know if we should wait for that before we review this PR. The benefit of using the extension is that we can manipulate the way things look any time that we want.


evaneschneider (approved):

evaneschneider:
This all looks good; gonna merge it in assuming that the failing Jenkins tests are a red herring.

PR #451, opened 2025-09-17
Port more documentation from the wiki

helenarichie:
Added the Particle-based Feedback, Static Gravity, Particles, and Cosmology pages to the Physics documentation. Particles and Cosmology are currently just placeholders. Also added 2D Rayleigh-Taylor to the documented examples.


evaneschneider:
Most of the information on the particle feedback page seems to be missing, is that intentional? Also I think the gravity page should just be called 'gravity' (as on the wiki) not 'static gravity', which is somewhat misleading. Otherwise this looks good.

evaneschneider:
FYI, I'm merging this without the requested edits so that Matthew can incorporate the particle feedback in PR #386

PR #452, opened 2025-09-18
Clean up Jenkins log

mabruzzo:
As I understand it, the Jenkins `sh` step (for executing shell scripts) sets the interpretter based on the presences of a shebang. When no shebang is present, it defaults to using the system default shell with the `-xe` flag. For the uninitiated,
* `-x` enables tracing mode
* `-e` tells the script to exit on errors

The reference page can be found here:
https://www.jenkins.io/doc/pipeline/steps/workflow-durable-task-step/#sh-shell-script

Ordinarily, setting the `-x` flag is a really nice choice. But, because the shell configuration script does so much work on the CRC cluster, it ends up polluting the logs.

Consequently, this commit tries to specify a shebang to launch the shell shell scripts without enabling tracing mode.


evaneschneider:
Should this be closed now that #449 is merged?

mabruzzo:
Uh, this is doing something slightly different. This is trying to reduce the amount of "garbage" printed in the log. I need to take another look at this (and resolve all of the newly created conflicts)

PR #453, opened 2025-09-18
Refactor Rotated Projection Logic

mabruzzo:
This PR should be reviewed after #433 has been merged.

## Overview

This PR essentially makes 3 changes:
1. It introduces the `RotatedProjWriter` class. This is a callable (e.g. it acts like a function with state) that is tracked within `WriterManager`.
2. It does some light refactoring so that an instance of the `Rotation` struct is no longer tracked by the `Grid3D` class. Now `RotatedProjWriter` tracks an instance.
3. It refactors so that `Rotation` instances are no longer initialized from values in a `Parameters` struct. Now, values are directly parsed in the constructor of `Rotation`

## Review Advice

It may be easier to review this commit-by-commit. 

- The first four commits simply move around logic
- The fifth commit starts using the new parameter parsing approach
- The sixth commit moves the contents of `Ouptut_Rotated_Projection_Data` into the callable method of the `RotatedProjWriter` class. (This may be a little controversial since I am moving contents between files).
- The last commit does a little cleanup


src/io/RotatedProjWriter.cpp
```diff
@@ -66,13 +72,13 @@ void io::RotatedProjWriter::operator()(Grid3D &G, Parameters P, int nfile, const
   // create the filename
   std::string filename = fname_template.format_fname(nfile, "_rot_proj");
 
+  // it may be a little more explicit to use a switch statement instead of if/elif/else
```
helenarichie, line 83: I agree that a switch statement would be clearer if you wanted to change it (but I don't think it's necessary for this PR if you don't feel like it).

src/io/RotatedProjWriter.cpp
```diff
@@ -22,6 +23,10 @@ io::Rotation::Rotation(ParameterMap &pmap)
   // - delta, phi, & theta satisfy 0 <= x < 180 (maybe we want to be more flexible)
   // - Lx & Lz are positive (or at least non-zero)
   // - I think flag_delta should only be 0, 1, or 2
+  //
+  // should we make it an error to specify the:
+  // - ddelta_dt parameter when flag_delta != 1
+  // - n_delta parameter when flag_delta != 2
```
helenarichie, line 30: I do think it's a good idea to check that the values of these parameters are valid, at least for the basic things of being positive/nonzero (either in this PR or a future one).

Also, I think it's okay for the `n_delta`/`ddelta_dt` parameters to go unused if they're defined with the wrong `flag_delta` value. But if you're concerned about it being confused, I think printing a warning could also be a good option.

helenarichie (approved):
I think this looks great and is good to go as-is!
I shared my thoughts on a couple of the questions you left in the comments. If you want to make any of those changes in this PR, go ahead, and I'll review them, but if not, feel free to just resolve my comments.

PR #454, opened 2025-09-22
Basics of Writing a Cholla Param File Doc Page

mabruzzo:
I think this is ready to go


mabruzzo:
pre-commit.ci run

evaneschneider:
This looks good to me, but I do have one quick question - in the example for the tables, all of the parameters have the same key, 'my_param'. Is that intentional, to show that you can have parameters with the same name if they are part of different tables?

mabruzzo:
> Is that intentional, to show that you can have parameters with the same name if they are part of different tables?

Yup! that's exactly the point of it

PR #465, opened 2025-12-17
Add particle loading to `cholla_utils`

mabruzzo:
## Overview

This PR does a lot (and unfortunately, I wasn't very careful with structuring the commit history). This whole undertaking was a surprising amount of work

It does several (related) things:
1. It introduces `cholla_utils.load_particle_data` & `cholla_utils.get_native_ptype_properties`
2. It introduces new documentation for these functions
3. It overhauls the testing machinery. We now introduce lots and lots of logic for simulating snapshots in different formats (distributed vs legacy concatenation vs modern concatenation), and with or without particle data. This has allowed us to create a pretty rigorous test-suite
4. I refactored all existing tests to make use of this new machinery. In the process, I have repurposed some old logic to explicitly test the field-concatenation logic from our field-concatenation scripts (we can do similar things for testing our particle concatenation script and snapshot-repacking - but that's beyond the scope of this PR)
5. I added a new page to the documentation explaining how to actually run the `cholla_utils` tests

A lot of this functionality originated from code I wrote for the yt-frontend. But along the way (with all the rigorous testing), I spotted and fixed/improved a bunch of functionality.  (now I need to cross-pollinate back to the yt-frontend)

Let me know what you think. I'm happy to make changes.

## Walking through the new API for reading Particle Data

(this is adapted from the new docs)

Let's walk through a few scenarios related to particle data.
For concreteness, suppose that the particle data for a snapshot is saved at **path/to/snap/7_particles.h5.0**.

> **Important aside:** similar to the functionality for reading field data, all of these examples will work if:
> - **path/to/7_particles.h5.0** contains the entire concatenated snapshot
>   - (yes, I know standard naming convention dictates that a snapshot holding concatenated data is typically called **7_particles.h5** rather than **7_particles.h5.0**)
>   - Unlike with the field data, this functionality can **ONLY** read concatenated particle data created by modern concatenation scripts (i.e. the resulting file maintains details about the original block-structure).[^1]
> - OR, if **path/to/snap.h5.0** is one of many files that the snapshot's particle data is distributed between (i.e. there is no concatenation).
>   - This is the default format that Cholla writes, and the data is split between files that share the same path, except for the suffix after **.h5**
>   - In this case, it **is** important that you use the path variant that ends in **.0**.

Let's move onto the examples:

1. If you want to query the particle properties that are saved within the dataset, you can invoke the following snippet:

   ```python
   import cholla_utils

   prop_pairs = cholla_utils.get_native_ptype_properties(
       "path/to/7_particles.h5.0"
   )
   print(prop_pairs)  # prints out all particle-property pairs
   ```

   The result is a list of two-tuples of the form `(<ptype>, <prop>)`, where `<ptype>` specifies particle-type and `<prop>` specifies a property name.
   This obviously merits further elaboration:
   - the choice to identify quantities associated with particles by these pairs, rather than just by the property name alone, was made to future-proof the API (and borrows from yt-conventions).
   - At the time of writing, Cholla is only capable of modelling (at most) a single particle type in a simulation.
     However, it's not hard to imagine a future where we add support for multiple types of particles (e.g. star-particles, dark matter particles, tracer particles), that may have different sets of properties (e.g. we might care about tracking masses and creation times for star particles, but not for dark matter particles).
   - At the time of writing, Cholla's distributed files don't directly record a particle type.
     We adopt yt's conventions and refer to these particles as having the `"io"` type.
     Cholla's particle concatenation script allows the user to specify an arbitrary particle name, but it too defaults to the `"io"` datatype.
   - standard particle properties include `"particle_IDs"`, `"pos_x"`, `"pos_y"`, ...

6. Loading a single field: This is easy. If that snapshot has a property called `("io", "pos_x")`, you can simply invoke:

   ```python
   import cholla_utils

   pos_x = cholla_utils.load_particle(
       "path/to/7_particles.h5.0", ("io", "pos_x")
   )
   ```

   In the above snippet, the `pos_x` variable holds a 1D numpy array.

   Imagine that we are only interested in working with a subset of the particle data.
   For concreteness, imagine that we only want to consider the data associated with the block in the leftmost corner of the simulation.
   The following snippet illustrates how we might accomplish this:

   ```python
   import cholla_utils

   density = cholla_utils.load(
       "path/to/7_particles.h5.0", ("io", "pos_x"), block_idx=np.s_[0,0,0]
   )
   ```

   If we instead wanted all blocks along the right x boundary, we could swap instead pass `np.s_[-1,:,:]` as the `block_idx` keyword argument.

7. Loading multiple particle properties: if we want to load the `("io", "particle_IDs")` and `("io", "pos_x")` particle properties at the same time, we would invoke:

   ```python
   import cholla_utils

   pos_x = cholla_utils.load_particle(
       "path/to/7_particles.h5.0", [("io", "particle_IDs"), ("io", "pos_x")]
   )
   # data is a dict
   # data[("io", "particle_IDs")] holds the particle_IDs numpy array
   # data[("io", "pos_x")] holds the pos_x numpy array
   ```

   Once again, we can tell the load function to only load a subset of the data by passing the `block_idx` keyword argument.

## Useful filename handling logic when analyzing both Particles and Field Data

When Cholla writes snapshots, it typically writes field data to a file named something like `63.h5.0` and particle data to a file of `63_particles.h5.0`.
If the files following this naming convention, the package provides convenience functionality such that `cholla_utils.load_particle` and `cholla_utils.get_native_ptype_properties` will automaticlly infer that the user wants to read `63_particles.h5.0` if the functions are passed `63.h5.0`.

Furthermore, this functionality also works if:
- the field data is concatenated and the particle data is distributed
- the field data is distributed and the particle data is concatenated
- the field data and the distributed data are both concatenated.

[^1]: Cholla historically provided legacy particle concatenation scripts that mirrored the behavior of the legacy field concatenation scripts.
      However, we have decided not to support this format because, to our knowledge, these scripts were never used to concatenate production datasets.


python/README.md
```diff
@@ -92,13 +99,104 @@ Onto the examples:
...
+
+## Helpful Tips For Analyzing Both Particles and Field Data
+
+When Cholla writes snapshots, it typically writes field data to a file named something like `63.h5.0` and particle data to a file of `63_particles.h5.0`.
+If the files following this naming convention, the package provides convenience functionality so that if `cholla_utils.load_particle` and `cholla_utils.get_native_ptype_properties` will automaticlly infer that the user wants to read `63_particles.h5.0` if the functions are passed `63.h5.0`.
+
+Furthermore, this functionality also works if:
+- the field data is concatenated and the particle data is distributed
+- the field data is distributed and the particle data is concatenated
+- the field data and the distributed data are both concatenated.
```
evaneschneider: typo: should be "the field data and the particle data" (I think)

evaneschneider (commented):
I didn't go over everything with a fine-toothed comb, but I think this looks great. The synthetic dataset generation stuff seems like it could be particularly useful if we want to add more (testable) analysis examples in the future.

python/README.md
mabruzzo: good catch!

PR #468, opened 2026-01-28
Refactor `Read_Grid_HDF5` to make use of `FieldInfo`

mabruzzo:
This PR is a followup to #467

Quite simply, this PR refactors `Read_Grid_HDF5` to make use of the newly introduced `FieldInfo` type.

### DIFF
--- a/src/io/io.cpp
+++ b/src/io/io.cpp
@@ -10,6 +10,7 @@
 #include <fstream>
 #include <iostream>
 #include <string>
+#include <unordered_set>
 #ifdef HDF5
   #include <hdf5.h>
 #endif  // HDF5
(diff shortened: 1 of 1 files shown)


src/io/io.cpp
```diff
@@ -2712,6 +2630,38 @@ void Grid3D::Print_Grid_Stats(void)
...
+        HII_density[id]   = INITIAL_FRACTION_HII * density[id];
+        HeI_density[id]   = INITIAL_FRACTION_HEI * density[id];
+        HeII_density[id]  = INITIAL_FRACTION_HEII * density[id];
+        HeIII_density[id] = INITIAL_FRACTION_HEIII * density[id];
+        e_density[id]     = INITIAL_FRACTION_ELECTRON * density[id];
+      }
+    }
+  }
+}
+
```
mabruzzo, line 2564: The following snippet is more consistent with the original style of the code that we replaced. I'm happy to revert it if that's your preference. But, be aware, I'm 99% that the old version is incompatible with Grackle.

For context, the `defined(COOLING_GRACKLE)` condition has been inserted for historical consistency, but I'm pretty sure the code won't compile -- `Grid3D::Conserved` doesn't have data members named `HI_density`, `HII_density`, ... in that mode

```C++
// this should work whether Cholla is configured with COOLING_GRACKLE or CHEMISTRY_GPU
// - earlier versions of this logic wouldn't work with Grackle
static void cosmo_init_chemical_species_(const Header &H, Grid3D::Conserved& C)
{
  #if defined(COOLING_GRACKLE) || defined(CHEMISTRY_GPU)
  for (int k = 0; k < H.nz_real; k++) {
    for (int j = 0; j < H.ny_real; j++) {
      for (int i = 0; i < H.nx_real; i++) {
        int id            = (i + H.n_ghost) + (j + H.n_ghost) * H.nx + (k + H.n_ghost) * H.nx * H.ny;
        int buf_id        = k + j * (H.nz_real) + i * (H.nz_real) * (H.ny_real);
        C.HI_density[id]    = INITIAL_FRACTION_HI * C.density[id];
        C.HII_density[id]   = INITIAL_FRACTION_HII * C.density[id];
        C.HeI_density[id]   = INITIAL_FRACTION_HEI * C.density[id];
        C.HeII_density[id]  = INITIAL_FRACTION_HEII * C.density[id];
        C.HeIII_density[id] = INITIAL_FRACTION_HEIII * C.density[id];
        C.e_density[id]     = INITIAL_FRACTION_ELECTRON * C.density[id];
      }
    }
  }
  #endif
}
```

helenarichie: I agree that it looks like the original version of this code would not work with `COOLING_GRACKLE`. This fix looks good to me!

helenarichie (approved):
Looks good to me!

PR #469, opened 2026-02-03
Refactor `Output_float32`

mabruzzo:
To be reviewed after #477 is merged

-------

In short this aims to make use of the new `FieldInfo` type and refactor parsing of Float32 parameters. The `Output_float32` logic is now implemented by the `F32FieldWriter` class.

I have factored out a bunch of common functionality between `F32FieldWriter` and `FieldWriter`. I have gone to great lengths to retain historical behavior. Once we reconcile all of the inconsistencies, a bunch of duplicated code can be removed


src/io/FieldWriter.h
```diff
@@ -5,40 +5,69 @@
...
+      : field_id{field_id}, name{std::move(name)}, io_buf{io_buf}, condition{condition}
+  {
+  }
+#endif
 };
 
+/*! Temporary type for tracking field-writer configuration
+ *
+ *  \note
+ *  In PR#469, we'll store the members of this struct directly within @ref FieldWriter
```
evaneschneider: This is PR#469, so is this comment out of date?

evaneschneider (approved):
Only one minor comment, otherwise I think this looks good.

PR #471, opened 2026-02-05
Adiabatic expansion test ICs and documentation

brantr:
I've created a simple adiabatic expansion test for an expanding cosmological frame. It simply populates a grid with the critical density in baryons and evolves the density and temperature adiabatically. There is no initial velocity in the gas, and no particle distribution.

I created a RTD entry for the 3D Adiabatic Expansion test. I confirm that the test runs successfully using dev.


brantr:
pre-commit.ci autofix

src/grid/initial_conditions.cpp
```diff
@@ -1557,6 +1559,79 @@ void Grid3D::Zeldovich_Pancake(struct Parameters P)
         E    = ics_values[2 * nPoints + index];
         U    = ics_values[3 * nPoints + index];
         // //
+        // temp = T_init * pow( dens/rho_0, 2./3.);
+        // U = temp / (gamma-1) / MP * KB * 1e-10 * dens;
+        // E = 0.5*dens*vel*vel + U;
+
```
evaneschneider, line 1565: Are these additions in the Zeldovich Pancake ICs intentional?

evaneschneider: (It's possible git is just confused about which parts of the file have changed, given that below it appears as though the end of the new Adiabatic Expansion test has not been altered, I just want to double-check that this file looks as intended.)

src/particles/particles_3D.cpp
```diff
@@ -214,6 +214,8 @@ void Particles3D::Initialize(Parameters *P, const SpatialDomainProps &spatial_pr
     Initialize_Sphere(P);
   } else if (strcmp(P->init, "Zeldovich_Pancake") == 0) {
     Initialize_Zeldovich_Pancake(P);
+  } else if (strcmp(P->init, "Adiabatic_Expansion") == 0) {
```
evaneschneider, line 217: I assume these modifications to the particles ICs are only necessary because the cosmology make-type by default includes particles?

evaneschneider:
CI tests are running now, but this looks good to merge once everything passes. (I have to trigger the tests manually for PRs outside of Pitt, since they run on the CRC hardware.) Thanks very much for adding the documentation for the test! Let me know if you'd like us to add it as a system test in the future - right now the only cosmology system test we have is a full cosmological simulation that as far as I know has never been verified.

src/grid/initial_conditions.cpp
brantr: Yes, they are intentional.  They reflect how the adiabatic ICs are set.  You can remove them if you want, and I can add them back after I am done with testing the Zeldovich test.

src/particles/particles_3D.cpp
brantr: Yes, in tests I tried to compile the code with cosmology and without particles and the compilation failed.

brantr:
Thanks. I plan to revise and test the cosmological simulation test as well.  We can add as a system test in the future, but I may revise this test depending on the outcome of the Zeldovich test (I may add a bulk velocity to check its evolution with scale factor).

PR #475, opened 2026-02-17
Refactor text outputs

mabruzzo:
To be reviewed after #473 is merged

-------

This does 3 things:

1. it removes support for making text outputs of multi-dimensional simulations (as we previously discussed)
2. it rewrites the text-field data outputs in order to make use of the `FieldInfo` class
3. the output text-field data now includes data for each of the passive scalars

Additional refactoring is done in #477 (that was split off from this PR to make things easier to review)

------

I validated the outputs in comparisons with older versions of the code.


src/io/FieldWriter.cpp
```diff
@@ -0,0 +1,220 @@
...
+  std::fputc('\n', fp);
+
+  // write all normal rows
+  for (i = H.n_ghost; i < H.nx - H.n_ghost; i++) {
+    std::fprintf(fp, "%d", i - H.n_ghost);
+    for (int field_idx = 0; i < n_fields; i++) {
+      std::fprintf(fp, "\t%f", ptr_arr[field_idx][i]);
+    }
+    std::fputc('\n', fp);
+  }
```
mabruzzo, line 152: The repeated calls to `fprintf` to write every individual field in a row is probably pretty slow.

A faster alternative would be to:
- pre-allocate a string buffer outside of this loop
- replace each fprintf and the fputc call with operations to write all the values to the string buffer
- and then call `fputs` (or possibly `fwrite`) at the end of the loop to write the entirety of the string buffer to disk all at once.

In fact, I got extremely far with writing an implementation that does exactly this. I made use of [`std::to_chars`](https://en.cppreference.com/w/cpp/utility/to_chars.html) to make things even faster. However, things got quite complicated and I ultimately decided to go with a much simpler (and much slower alternative).

If you want me to post the alternative code, just let me know.

evaneschneider (approved):

PR #477, opened 2026-02-26
Unify text-outputs and hdf5-outputs

mabruzzo:
To be reviewed after #475 is merged

This was originally going to be part of #475, but I decided to handle it separately to make the individual changes easier to understand.

---------

### Primary Change

The primary focus of this PR is to unify configuration of the fields that actually get written out during a data dump.

Prior to this PR:
- the choice of fields written to native resolution hdf5 fields was controlled by compiler flags
- the choice of fields written to text files was fixed and totally unaffected by those compiler flags

After this PR: The logic for determining the fields written to disk is now unified. In other words, this logic doesn't care whether we write native resolution hdf5 files or text files.

Consequences:
- Technically, the choice of serialized fields remains controlled by compile-time options. But the changes here won't hinder the process of moving that choice to runtime parameters (and if anything that choice will become increasingly useful)
- With the changes made here, it wouldn't be hard to extend the codebase to make Cholla support runtime parameters for writing both hdf5 and textfiles in the same simulation

> [!note]
> The preceding discussion applies to all fields other than gravitational potential and temperature (those are special cases that can only be written as part of native-resolution-hdf5 files).

### Secondary Changes

This PR also made 2 other changes:
- the program provides a warning at startup if the simulation isn't 1D, and the codebase is configured to write text files
- the column names in text files for magnetic fields were changed from `magX`, `magY`, `magZ` to `magnetic_x`, `magnetic_y`, `magnetic_z` (i.e. we now directly use the field names). This won't cause any backwards-compatability issues since Cholla can't really run 1D MHD simulations

### Followups

A bunch of this work serves as a foundation for the refactored 32bit HDF5 output logic

------

I have manually validated the correctness of the output files

### DIFF
--- a/src/io/FieldWriter.cpp
+++ b/src/io/FieldWriter.cpp
@@ -45,7 +45,7 @@ namespace
  *  - once we finish constructing @ref FieldWriter is constructed, the instance of this
  *    type is discarded
  */
-struct DsetSpecListBuilder_ {
+struct DsetSpecListBuilder {
   using OutNameRecipie = std::function<std::string(std::string_view)>;
 
  private:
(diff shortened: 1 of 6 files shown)


src/io/FieldWriter.cpp
```diff
@@ -21,6 +22,105 @@
...
+struct DsetSpecListBuilder {
+  using OutNameRecipie = std::function<std::string(std::string_view)>;
+
+ private:
+  /*! reference to the vector that the instance gradually "builds-up" */
+  std::vector<io::DatasetSpecEntry>& wrapped_vec;
+  /*! references the field_info from the current simulation. */
+  const FieldInfo& field_info;
+  /*! returns the output name used for a field based upon the original field name */
+  OutNameRecipie out_name_recipe;
```
evaneschneider, line 57: Probably want this to be OutNameRecipe

PR #479, opened 2026-03-03
Refactor projection logic

mabruzzo:
To be reviewed after #469 is merged

---------

This refactors the projection-writing functionality.

The goal here was to make the structure of the projection writing logic more closely resemble the logic of other output writers. Beyond consistency, there isn't much other immediate benefit (longer term, the changes make it easier to pass configuration parameters to the writer).

I did some light refactoring, but didn't bother to make use of `FieldInfo` (at this moment, the would be somewhat marginal -- but if we added support for customizing which fields are projected via runtime parameters OR projected more fields, it would make sense to use `FieldInfo`)

NOTE: the projected temperature values changed ever so slightly. But that's entirely due to compiler optimizations (I took a lot of care to ensure I didn't break any logic)


src/io/ProjectionWriter.h
```diff
@@ -0,0 +1,50 @@
...
+#include "../grid/grid3D.h"
+
+struct ParameterMap;
+struct FieldInfo;
+struct FnameTemplate;
+
+namespace io
+{
+
+/*! \brief A callable object that writes slice data
```
evaneschneider: "projection" data

src/io/ProjectionWriter.h
```diff
@@ -0,0 +1,50 @@
...
+ *  At the time of writing, this doesn't get configured at startup and doesn't need to
+ *  a full-blown class. Right now, this mostly exists for consistency with other output
+ *  approaches.
+ */
+class ProjectionWriter
+{
+ public:
+  ProjectionWriter() = default;
+
+  /*! Implements the a callable method that writes slice data
```
evaneschneider: "projection" data

evaneschneider (approved):

PR #481, opened 2026-03-17
Introduce `H5AttrWriter`

mabruzzo:
To be reviewed after #469 is merged

--------

This PR introduces the `H5AttrWriter` type. This is a temporary type that wraps a `hid_t` that wraps an hdf5 file (or an hdf5 group). The intention is to provide a nicer interface for writing hdf5 attributes to disk.

This was initially intended to be short and quick, but after making a bunch of progress, I hit some subtle and very annoying bugs that took a surprisingly long time to debug. The resulting code is a little longer than I would like.

A side motivation: in a followup PR, we make this class inherit from an abstract `AttrWriterAdaptor` class so that we can unify the code for text headers and H5 headers

-------

I have tested to confirm that the headers are identical to the earlier version of the code


src/io/io.cpp
```diff
@@ -460,20 +350,94 @@ void Grid3D::Write_Header_Rotated_HDF5(hid_t file_id, io::Rotation &R)
...
+  }
+  if (H5Aclose(attribute_id) < 0) {
+    CHOLLA_ERROR("error closing the \"%s\" attribute", name);
+  }
+}
+
+void H5AttrRecorder::record_arr(const char *name, const double *arr, int length)
+{
+  hid_t type_id = H5T_NATIVE_DOUBLE;
+  // is there a compelling reason why this is big endian?
```
evaneschneider, line 401: No, in fact there isn't and it has caused problems in the past. Perhaps we should change it?

evaneschneider (approved):

PR #482, opened 2026-03-17
Refactor `RotatedProjWriter`

mabruzzo:
To be reviewed after #481 is merged

--------

This PR refactors a bunch of logic so all the projection-related logic for `RotatedProjWriter` is within **RotatedProjWriter.cpp**.

As part of this PR, I factored out the common logic for writing the common header attributes present in the rotated projection outputs and the headers of projections, slices, and full field dumps. I have taken a lot of care to ensure that I don't introduce new headers.

Would it be ok if I add all the header attributes from projections, slices, and full field dumps to rotated projection dumps? For context, the attributes are the same for those 3 output types. (From a code perspective, it simplifies logic. Plus, I can't think of a compelling reason why we would want Rotated Projections to be different)

-----
I did a lot of careful testing to make sure I didn't break anything (which took a surprising amount of time)


evaneschneider:
These changes look good, but io needs a refactor since I merged the slice PR in first.

src/io/RotatedProjWriter.cpp
```diff
@@ -13,7 +13,303 @@
...
+
+#ifdef HDF5
+
+/*! \brief Write rotated projected data to a file, at the current simulation */
+void Write_Rotated_Projection_HDF5_(const Grid3D &G, hid_t file_id, const io::Rotation &R)
+{
+  const Header &H = G.H;
+  bool is_3D      = H.nx > 1 && H.ny > 1 && H.nz > 1;
+  if (not is_3D) {
+    chprintf("Slice write only works for 3D data.\n");
```
evaneschneider:
```suggestion
    chprintf("Projection writes only work for 3D data.\n");
```

src/io/RotatedProjWriter.h
```diff
@@ -82,7 +87,13 @@ class RotatedProjWriter
   RotatedProjWriter() = delete;
   RotatedProjWriter(ParameterMap &pmap) : rot_info_(pmap) {}
 
-  /*! A callable method that writes a rotated projection of the grid data to file.
+  /*! Writes the rotated project to disk.
```
evaneschneider:
```suggestion
  /*! Writes the rotated projection to disk.
```

src/io/RotatedProjWriter.cpp
```diff
@@ -13,7 +13,303 @@
...
+  a22 = ct;
+
+  *xp = a00 * x + a01 * y + a02 * z;
+  *yp = a10 * x + a11 * y + a12 * z;
+  *zp = a20 * x + a21 * y + a22 * z;
+}
+
+#ifdef HDF5
+
+/*! \brief Write rotated projected data to a file, at the current simulation */
```
evaneschneider:
```suggestion
/*! \brief Write rotated projected data to a file, at the current simulation time */
```

evaneschneider (approved):

PR #485, opened 2026-03-18
Some cleanup of unused build-time configuration options

mabruzzo:
A followup to #483

-------

Back when I wrote #483, I used `grep` and regex commands to extract every single macro referenced by the builds directory. 8 of them were spurious options that never get used for anything (in just about every case they showed up in commented lines). This PR cleans up these unused options


builds/make.type.dust
```diff
@@ -27,7 +27,7 @@ DFLAGS    += -DDUST
 
 # Apply the cooling in the GPU from precomputed tables
 DFLAGS    += -DCOOLING_GPU
-DFLAGS    += -DCLOUDY_COOLING
+DFLAGS    += -DCLOUDY_COOL
```
mabruzzo, line 30: It's important to note that because `CLOUDY_COOLING` was a typo for `CLOUDY_COOL`, this change fundamentally changes the default choice of the cooling function.

- Previous default: when `COOLING_GPU` is defined without `CLOUDY_COOL`, we used piecewise-cie
- New default: when `COOLING_GPU` and `CLOUDY_COOL` are both defined, we default to the cloudy cooling choice

Now, we *could* just entirely delete the highlighted line to go back to the old default. But none of this really matters (for testing purposes) since we don't run any full-blown dust simulations. Furthermore, #502 will entirely remove the macros used for setting the default cooling function.

PR #486, opened 2026-03-23
Introduce `AttrRecorderInterface` and `TextAttrRecorder`

mabruzzo:
This is a followup to #481 (but builds on some changes from #482)

-----

Essentially, this PR:
- introduces the `AttrRecorderInterface` abstract base class
- modifies `H5AttrRecorder` so that it's a subclass of `AttrRecorderInterface`
- introduces `TextAttrRecorder`, a class that implements the interface of `AttrRecorderInterface` and writes header attributes in TOML format.
- The logic for recording the header information has been refactored so that the logic all uses the `AttrRecorderInterface` interface (in other words, equivalent header information is written to both hdf5 or text files)


helenarichie (approved):
This is really nice - looks ready to merge to me!

PR #495, opened 2026-03-30
Remove final parameters that were parsed with the old style of parsing

mabruzzo:
To be reviewed after #494 has been merged

-------

This gets rid of the final old-style parsing logic. It does a little associated cleanup.

### DIFF
--- a/src/global/global.cpp
+++ b/src/global/global.cpp
@@ -136,36 +136,7 @@ void Warn_Unused_Params(ParameterMap &pmap) { pmap.warn_unused_parameters(option
  *
  *  \returns true if the parameter was actually used. false otherwise.
  */
-bool Old_Style_Parse_Param(const char *name, const char *value, struct Parameters *parms)
-{
-  /* Copy into correct entry in parameters struct */
-  if (strcmp(name, "output_always") == 0) {
-    int tmp = atoi(value);
-    // In this case the CHOLLA_ASSERT macro runs into issuse with the readability-simplify-boolean-expr clang-tidy check
-    // due to some weird macro expansion stuff. That check has been disabled here for now but in clang-tidy 18 the
-    // IgnoreMacro option should be used instead.
-    // NOLINTNEXTLINE(readability-simplify-boolean-expr)
-    CHOLLA_ASSERT((tmp == 0) or (tmp == 1), "output_always must be 1 or 0.");
-    parms->output_always = tmp;
-  } else if (strcmp(name, "n_steps_limit") == 0) {
-    parms->n_steps_limit = atof(value);
-#ifdef PARTICLES
-  } else if (strcmp(name, "prng_seed") == 0) {
-    parms->prng_seed = atoi(value);
-#endif  // PARTICLES
-  } else if (strcmp(name, "bc_potential_type") == 0) {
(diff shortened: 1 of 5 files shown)


src/global/global.cpp
```diff
@@ -327,6 +283,29 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+    parms->output_always = tmp;
+  }
+
+  parms->n_steps_limit = pmap.value_or("n_steps_limit", -1);
+
+#ifdef PARTICLES
+  parms->prng_seed = pmap.value_or("prng_seed", 0);
+#endif  // PARTICLES
+
+#ifdef GRAVITY
```
evaneschneider: It seems like this didn't used to be wrapped in the GRAVITY macro. Perhaps that is the source of the failing gravity and cosmology system tests.

mabruzzo: Good catch! That's probably the issue

src/global/global.cpp
```diff
@@ -327,6 +283,28 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+  parms->prng_seed = pmap.value_or("prng_seed", 0);
+#endif  // PARTICLES
+
+  // a negative value means that the parameter wasn't set
+  parms->bc_potential_type = pmap.value_or("bc_potential_type", -1);
+
+#if defined(SCALAR) && defined(DUST)
+  // we are assuming that this is an integer for historical consistency... But it sure
+  // seems like this should be a double
+  parms->grain_radius = pmap.value<int>("grain_radius");
```
evaneschneider: @helenarichie Can you confirm whether this should be an integer?

helenarichie: You're right, that should definitely be a double! Thanks for catching that.

evaneschneider:
```suggestion
  parms->grain_radius = pmap.value<double>("grain_radius");
```

src/global/global.cpp
```diff
@@ -327,6 +283,28 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+
+#ifdef PARTICLES
+  parms->prng_seed = pmap.value_or("prng_seed", 0);
+#endif  // PARTICLES
+
+  // a negative value means that the parameter wasn't set
+  parms->bc_potential_type = pmap.value_or("bc_potential_type", -1);
+
+#if defined(SCALAR) && defined(DUST)
+  // we are assuming that this is an integer for historical consistency... But it sure
```
evaneschneider:
```suggestion
```

src/global/global.cpp
```diff
@@ -327,6 +283,27 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+
+#ifdef PARTICLES
+  parms->prng_seed = pmap.value_or("prng_seed", 0);
+#endif  // PARTICLES
+
+  // a negative value means that the parameter wasn't set
+  parms->bc_potential_type = pmap.value_or("bc_potential_type", -1);
+
+#if defined(SCALAR) && defined(DUST)
+  // seems like this should be a double
```
evaneschneider:
```suggestion
```

PR #497, opened 2026-03-30
Convert `Init_Param_Struct_Members` & `Parse_Params` into a Constructor for `Parameters`

mabruzzo:
To be reviewed after #495 is merged

--------

This PR does some basic cleanup now that `Old_Style_Parse_Param` was deleted within #495.

After deleting `Old_Style_Parse_Param`, there isn't much reason for `Init_Param_Struct_Members` & `Parse_Params` to be separate functions (the former now simply calls the latter). Plus, I thought it made sense to make the combined function into a constructor for `Parameters` (since that is all the function actually does)


src/global/global.cpp
```diff
@@ -327,6 +268,28 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
...
+  parms->prng_seed = pmap.value_or("prng_seed", 0);
+#endif  // PARTICLES
+
+  // a negative value means that the parameter wasn't set
+  parms->bc_potential_type = pmap.value_or("bc_potential_type", -1);
+
+#if defined(SCALAR) && defined(DUST)
+  // we are assuming that this is an integer for historical consistency... But it sure
+  // seems like this should be a double
+  parms->grain_radius = pmap.value<int>("grain_radius");
```
helenarichie: Didn't this get changed to a double in #495?

Regardless, it should be `parms->grain_radius = pmap.value<double>("grain_radius");` and the comment should be removed.

helenarichie (changes requested):
This looks good aside from the one issue with the dust radius parameter

helenarichie (approved):

src/global/global.cpp
mabruzzo: Ah, indeed it was changed. This commit was just built off of the branch from PR #495 before we fixed that.

now that @evaneschneider merged in the dev branch, this has been fixed

mabruzzo:
@evaneschneider, this is good to be merged whenever you are ready

PR #500, opened 2026-03-31
Cosmo dynamical de

astrodiegog:
Include ability to calculate the expansion history of cosmological simulations with a time-evolving dark energy equation of state

- Create new parameter expected as `wDE_file` which points to a file with z vs wDE(z) entries
- Create macro `DYNAMICAL_DE_TABLE` so that compiler knows whether to allocate certain variables
- Create routine that integrates wDE(z) to find rhoDE(z)/rhoDE(z=0) using midpoint rule
- Linearly interpolate between rhoDE(z)/rhoDE(z=0) entries for arbitrary redshifts
- In verifying the expansion history file output, realized the the expansion history was updated before output : `Cosmo.dt += dt_secs` occurred before `Cosmo.Write_Expansion_History_Entry()`. Updated so that expansion history entry occurs before updating the time
- Compilation with `DYNAMICAL_DE_TABLE` overrides `w0, wa` parameter file entries


astrodiegog:
pre-commit.ci autofix

astrodiegog:
@brantr Comment whenever you have the time!

evaneschneider:
I'll take a more thorough look at this tomorrow (we have a hack day in the afternoon), but a couple quick comments:
-Since the first CAAR project, we've been trying hard to decrease macro use, because they make it much harder to test code, and are easy to forget to define, or define incorrectly. That doesn't mean we won't allow new compiler macros, but I am trying to encourage input parameter usage as opposed to macros whenever possible. So any new macros should be justified (explain in the PR why you need a macro to accomplish this). That way, if we can help you rewrite the code to NOT use a macro, we will have the opportunity to do so.
-Now that we have read the docs integrated into the codebase, I am enforcing documentation with PRs that include new code capabilities. So this PR should be modified to include, at a minimum, a page (probably in the physics subdirectory) that explains the dynamical dark energy capability AND how it is implemented (the new macro needs to be documented somewhere, there should be a minimum description of the contents of the expected DE file...). I know this may seem silly since as of right now the cosmology portion of Cholla is virtually undocumented, but we have to start somewhere.

src/global/global.cpp
```diff
@@ -392,6 +392,12 @@ void Init_Param_Struct_Members(ParameterMap &pmap, struct Parameters *parms)
   parms->Omega_R       = pmap.value_or("Omega_R", 0.0);
   parms->w0            = pmap.value_or("w0", -1.0);
   parms->wa            = pmap.value_or("wa", 0.0);
+
+  #ifdef DYNAMICAL_DE_TABLE
+  // using macro to let compiler know whether to declare wDE table
+  Load_String_Param_Into_Char_Buffer(pmap, "wDE_file", parms->wDE_file, nullptr);
+  #endif
```
mabruzzo:
```suggestion
  // if the wDE table isn't provided, store an empty string
  Load_String_Param_Into_Char_Buffer(pmap, "wDE_file", parms->wDE_file, "");
```

src/cosmology/cosmology.h
```diff
@@ -23,6 +23,18 @@ class Cosmology
...
+  float *dynamicalDE_table_z;
+  float *dynamicalDE_table_w;
+  // dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  float *dynamicalDE_table_density;
+  int n_wDE_samples;
+  void Load_DynamicalDE_EquationOfState(struct Parameters *P);
+  Real Get_wDE_from_a(Real a);
+  void Set_DynamicalDE_Density();
+  Real Get_DynamicalDE_Density_from_a(Real a);
+    #endif
```
mabruzzo:
```suggestion
  float *dynamicalDE_table_z;
  float *dynamicalDE_table_w;
  // dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
  float *dynamicalDE_table_density;
  int n_wDE_samples;
  void Load_DynamicalDE_EquationOfState(struct Parameters *P);
  Real Get_wDE_from_a(Real a);
  void Set_DynamicalDE_Density();
  Real Get_DynamicalDE_Density_from_a(Real a);

  /*! Indicate whether the simulation is configured to use a DynamicalDE EOS table */
  bool Using_DynamicalDE_Table() { return n_wDE_samples != 0; }
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +24,130 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
   chprintf("Cosmology Successfully Initialized. \n\n");
 }
 
+  #ifdef DYNAMICAL_DE_TABLE
+
```
mabruzzo:
```suggestion
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +24,130 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
+    prev_integral_sum = integral_sum_arr[i - 1];
+    integral          = prev_integral_sum + my_integral;
+
+    integral_sum_arr[i] = integral;
+
+    dynamicalDE_table_density[i] = onez_3 * exp(3. * integral);
+  }
+}
+
+  #endif  // DYNAMICAL_DE_TABLE
```
mabruzzo:
```suggestion
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +24,130 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
-  Real fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
-  Real a_dot  = sqrt(Omega_R / a2 + Omega_M / a + a2 * Omega_L * fac_de + Omega_K) * H0;
+  Real a2 = a * a;
+  Real fac_de;
+
+  #if DYNAMICAL_DE_TABLE
+  fac_de = Get_DynamicalDE_Density_from_a(a);
+  #else
+  fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
+  #endif  // DYNAMICAL_DE_TABLE
```
mabruzzo:
```suggestion
  if (Using_DynamicalDE_Table()) {
    fac_de = Get_DynamicalDE_Density_from_a(a);
  } else {
    fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
  }
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -60,9 +178,16 @@ Real Cosmology::Get_dt_from_da_rk(Real da, Real a)
...
-  Real fac_de = pow(current_a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
-  Real a_dot  = sqrt(Omega_R / a2 + Omega_M / current_a + a2 * Omega_L * fac_de + Omega_K) * H0;
+  Real a2 = current_a * current_a;
+  Real fac_de;
+
+  #if DYNAMICAL_DE_TABLE
+  fac_de = Get_DynamicalDE_Density_from_a(current_a);
+  #else
+  fac_de = pow(current_a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
+  #endif  // DYNAMICAL_DE_TABLE
```
mabruzzo:
```suggestion
  if (Using_DynamicalDE_Table()) {
    fac_de = Get_DynamicalDE_Density_from_a(current_a);
  } else {
    fac_de = pow(current_a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
  }
```

src/cosmology/cosmology_functions.cpp
```diff
@@ -83,10 +208,16 @@ Real Cosmology::Get_dt_from_da(Real da, Real a)
...
+  Real a2 = a * a;
+  Real a3 = a2 * a;
+  Real a4 = a2 * a2;
+  Real fac_de;
+
+  #if DYNAMICAL_DE_TABLE
+  fac_de = Get_DynamicalDE_Density_from_a(a);
+  #else
+  fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
+  #endif  // DYNAMICAL_DE_TABLE
```
mabruzzo:
```suggestion
  if (Using_DynamicalDE_Table()) {
    fac_de = Get_DynamicalDE_Density_from_a(a);
  } else {
    fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
  }
```

src/cosmology/cosmology.cpp
```diff
@@ -20,6 +20,10 @@ void Cosmology::Initialize(struct Parameters *P, Grav3D &Grav, Particles3D &Part
   Omega_b = P->Omega_b;
   w0      = P->w0;
   wa      = P->wa;
+  #ifdef DYNAMICAL_DE_TABLE
+  Load_DynamicalDE_EquationOfState(P);
+  Set_DynamicalDE_Density();
+  #endif
```
mabruzzo:
```suggestion
  
  // set default values pertaining to DynamicalDE EOS table
  dynamicalDE_table_z = nullptr;
  dynamicalDE_table_w = nullptr;
  dynamicalDE_table_density = nullptr;
  n_wDE_samples = 0;
  if (std::strlen(P.wDE_file) != 0) {
    Load_DynamicalDE_EquationOfState(P);
    Set_DynamicalDE_Density();
  }
```

src/cosmology/io_cosmology.cpp
```diff
@@ -87,4 +87,72 @@ void Cosmology::Set_Next_Scale_Output()
   }
 }
 
+  #ifdef DYNAMICAL_DE_TABLE
```
mabruzzo:
```suggestion
```

src/cosmology/io_cosmology.cpp
```diff
@@ -87,4 +87,72 @@ void Cosmology::Set_Next_Scale_Output()
...
+  chprintf("  z_min = %f    z_max = %f \n", dynamicalDE_table_z[0], dynamicalDE_table_z[n_wDE_samples - 1]);
+  chprintf("  w(z_min) = %f    w(z_max) = %f \n", dynamicalDE_table_w[0], dynamicalDE_table_w[n_wDE_samples - 1]);
+
+  if (dynamicalDE_table_z[0] != 0.) {
+    chprintf("We require z_min = 0 so that w(z=0) is well defined \n");
+    exit(1);
+  }
+}
+  #endif  // DYNAMICAL_DE_TABLE
+
```
mabruzzo:
```suggestion
```

src/global/global.h
```diff
@@ -290,6 +290,10 @@ struct Parameters {
   Real Init_redshift;
   Real End_redshift;
 
+  #ifdef DYNAMICAL_DE_TABLE
+  char wDE_file[MAXLEN];  // File with equation of state as function of redshift
+  #endif
```
mabruzzo:
```suggestion
  char wDE_file[MAXLEN];  // File with equation of state as function of redshift
```

builds/make.type.cosmology
```diff
@@ -29,3 +29,7 @@ DFLAGS += -DPRINT_INITIAL_STATS
 
 # Output all data every N_OUTPUT_COMPLETE snapshots ( These are Restart Files )
 DFLAGS += -DN_OUTPUT_COMPLETE=1
+
+# Solve Friedmann Equations with a given time-evolving dark energy equation of state table
+DFLAGS += -DDYNAMICAL_DE_TABLE
+
```
mabruzzo:
```suggestion
```

mabruzzo (commented):
(Feel free to ignore me) I took a quick look, and at a glance, it looks like it would be relatively straightforward to update this PR in order to get rid of the `DYNAMICAL_DE_TABLE` table.

The simplest way to get rid of the parameter is to update the code so that the `Parse_Params` function always checks (when the `COSMOLOGY` macro is defined) to see whether the `wDE_file` parameter is specified. If it's specified, then you will use. If it isn't specified, you could just store an empty string within ``Parameters::wDE_file``.

I included my recommendations for how you might do this down below. (I'm pretty confident it will "just work" if you go through and simply accept all the suggestions, but I haven't actually tested the code).

I do have a couple stylistic suggestions - but I'll hold off on those until after you decide on whether or not to remove the macro.

**EDIT:** OOPs! I realized that I forgot mention that this looks fantastic! It looks like a wonderful addition to Cholla!

evaneschneider:
On a related note, this PR fails the cosmology system test, because it changes the default cosmology build to one with dynamical dark energy. That's good! (Presumably, adding dynamical dark energy should change the results of a cosmological simulation!) But, you probably don't actually want to change the default cosmology system test. Simply removing the new macro from the cosmology build should fix this, but that's one of the downsides to approaching this change with a macro - if you wanted to add a new system test for dynamical dark energy (not a bad idea!), we'd have to add a whole new build type to the testing matrix. @mabruzzo made some suggestions for how you could accomplish this with an input parameter rather than a macro. Let us know if anything is confusing!

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +24,130 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
+void Cosmology::Set_DynamicalDE_Density()
+{
+  Real integrand_prev, integrand;
+  Real prev_integral_sum, my_integral, integral;
+  Real z_prev, onez_prev;
+  Real z, onez, onez_2, onez_3;
+  Real wDE_prev, wDE;
+
+  int i;
+  double integral_sum_arr[n_wDE_samples];
```
mabruzzo: I just wanted to highlight this before I forget: since `n_wDE_samples` isn't a compile-time constant, this line isn't valid C++ (it's a gcc extension). This *needs* to be changed before the PR can be merged.

You should replace this with 
```suggestion
  std::vector<double> integral_sum_arr(n_wDE_samples);
```

(you will also need to add `#include <vector>` at the top of this file

mabruzzo: This is a common mistake people make (there's a reason its a gcc extension)

mabruzzo:
> Now that we have read the docs integrated into the codebase, I am enforcing documentation with PRs that include new code capabilities. So this PR should be modified to include, at a minimum, a page (probably in the physics subdirectory) that explains the dynamical dark energy capability AND how it is implemented (the new macro needs to be documented somewhere, there should be a minimum description of the contents of the expected DE file...). I know this may seem silly since as of right now the cosmology portion of Cholla is virtually undocumented, but we have to start somewhere.

Just in case it helps, this page ([here](https://cholla.readthedocs.io/en/latest/Development/writing-docs.html)) should help you get started with adding docs.

Please reach out if have any questions or want help (with writing docs or the suggested changes); I'm more than happy to help!

astrodiegog:
> The simplest way to get rid of the parameter is to update the code so that the `Parse_Params` function always checks (when the `COSMOLOGY` macro is defined) to see whether the `wDE_file` parameter is specified. If it's specified, then you will use. If it isn't specified, you could just store an empty string within `Parameters::wDE_file`.
> 
> I included my recommendations for how you might do this down below. (I'm pretty confident it will "just work" if you go through and simply accept all the suggestions, but I haven't actually tested the code).

Sounds great! I committed your suggestions and am able to reproduce expansion history outputs. Thank you!

src/cosmology/cosmology_functions.cpp
astrodiegog: This makes sense, thank you for pointing this out to me!

src/cosmology/cosmology.h
```diff
@@ -23,6 +23,19 @@ class Cosmology
   Real w0;
   Real wa;
 
+  float *dynamicalDE_table_z;
+  float *dynamicalDE_table_w;
+  // dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  float *dynamicalDE_table_density;
+  int n_wDE_samples;
+  void Load_DynamicalDE_EquationOfState(struct Parameters *P);
+  Real Get_wDE_from_a(Real a);
```
mabruzzo: **IMPORTANT**: I'm 99% sure that this function is never used.

Why does it exist? Was it for testing? Can we delete it? If not, please add a docstring mentioning that the function isn't currently used and explaining why it exists.

src/cosmology/cosmology.h
```diff
@@ -23,6 +23,19 @@ class Cosmology
   Real w0;
   Real wa;
 
+  float *dynamicalDE_table_z;
+  float *dynamicalDE_table_w;
+  // dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  float *dynamicalDE_table_density;
+  int n_wDE_samples;
```
mabruzzo: **IMPORTANT**: you need to make a change relating to these data members to avoid memory leaks when the cosmology object is destroyed.

I would *STRONGLY* recommend that you convert each pointer to a `std::vector<float>`.
- the resulting declarations would look something like:
  ```c++
  std::vector<float> dynamicalDE_table_z;
  std::vector<float> dynamicalDE_table_w;
  // dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
  std::vector<float> dynamicalDE_table_density;
  ```
- you also have the option of deleting `n_wDE_samples` and everytime you would access that value, you would instead access `dynamicalDE_table_z.size()`

The alternative is to define a destructor for the `Cosmology` class. But, when a class tracks pointers, you generally need to give consideration to copy/move constructors and copy/move assignment operations. I'm happy to help with that, but I would generally discourage it.

mabruzzo: Yes, I know that a `Cosmology` object is usually destroyed when the simulation terminates so technically a memory leak doesn't affect the code when it runs. However, since it's bad programming practice to do that (especially since fixing it is so easy). More concretely, if we don't fix it has consequences:
- Pragmatically, memory debugging tools like valgrind (and presumably the memory sanitizer included with clang) will flag the lack of cleanup as an error (hindering the usefulness of such tools)
- If somebody eventually changed how exactly the `Cosmology` is constructed (or wrote any unit tests) they would be quite surprised with how it works

src/cosmology/cosmology.h
```diff
@@ -23,6 +23,19 @@ class Cosmology
...
+  void Load_DynamicalDE_EquationOfState(struct Parameters *P);
+  Real Get_wDE_from_a(Real a);
+  void Set_DynamicalDE_Density();
```
mabruzzo: **IMPORTANT:** This function is **only** called once, while initializing initializing the dynamical DE. I think it's fantastic that you made this into a separate method from `Load_DynamicalDE_EquationOfState` (and avoided making a single gigantic method).

Would you mind picking a name that makes it a little more obvious that it's only used during startup? Maybe something like `Setup_DynamicalDE_Density_`. Feel free to pick a different name you think is better.[^1]

It would also be great if you could add a short docstring explicitly mentioning that it is a setup function

[^1]: I personally like adding a trailing underscore to names that are only used internally (especially since we don't have any distinction between private and public methods). But that's just a personal preference. You totally don't have to do it.

src/cosmology/cosmology.cpp
```diff
@@ -20,6 +20,16 @@ void Cosmology::Initialize(struct Parameters *P, Grav3D &Grav, Particles3D &Part
   Omega_b = P->Omega_b;
   w0      = P->w0;
   wa      = P->wa;
+  
+  // set default values pertaining to DynamicalDE EOS table
+  dynamicalDE_table_z = nullptr;
+  dynamicalDE_table_w = nullptr;
+  dynamicalDE_table_density = nullptr;
```
mabruzzo: **IMPORTANT** If you convert the `dynamicalDE_table_*` data members so that they each have the type ``std::vector<float>`` (as I describe in a separate comment), you will need to simply delete these lines. 

If you choose not to do that, you should really move these lines so that they are executed within the constructor, ``Cosmology::Cosmology(void)`` (this is important for correctly implementing a destructor).

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +25,126 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
-  Real fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - current_a));
-  Real a_dot  = sqrt(Omega_R / a2 + Omega_M / a + a2 * Omega_L * fac_de + Omega_K) * H0;
+  Real a2 = a * a;
+  Real fac_de;
+
+  if (Using_DynamicalDE_Table()) {
+    fac_de = Get_DynamicalDE_Density_from_a(a);
+  } else {
+    fac_de = pow(a, -3 * (1 + w0 + wa)) * exp(-3 * wa * (1 - a));
+  }
```
mabruzzo: (Feel free to ignore!) This precise logic is repeated in 3 different places. Perhaps this logic should be converted into a method?

src/global/global.h
```diff
@@ -290,6 +290,8 @@ struct Parameters {
   Real Init_redshift;
   Real End_redshift;
 
+  char wDE_file[MAXLEN];  // File with equation of state as function of redshift
```
mabruzzo: (Definitely feel free to ignore!) I personally think it would be marginally better if this instead were

```C++
std::string wDE_file;  // File with equation of state as function of redshift
```

Doing that would require the following 2 changes:
1. line 397 of `src/global/global.cpp`: rather than call `Load_String_Param_Into_Char_Buffer` , the line would read be:
   ```C++
   parms->wDE_file = pmap.value_or("wDE_file", "");
   ```
2. line 92 of `src/cosmology/io_cosmology.cpp`: replace `if (P->wDE_file[0] == '\0') {` with `if (P->wDE_file.empty()) {`

src/cosmology/cosmology_functions.cpp
```diff
@@ -24,12 +25,126 @@ void Grid3D::Initialize_Cosmology(struct Parameters *P)
...
+  z_low   = dynamicalDE_table_z[i - 1];
+  wDE_low = dynamicalDE_table_w[i - 1];
+
+  lin_slope  = (wDE_low - wDE_high) / (z_low - z_high);
+  wDE_interp = lin_slope * (z - z_low) + wDE_low;
+
+  return wDE_interp;
+}
+
+/* Interpolate to get rhoDE(z) / rhoDE(z=0) */
```
mabruzzo:
```suggestion
/*! Interpolate to get rhoDE(z) / rhoDE(z=0) */
```

Adding the exclamation point is essential for making doxygen (our C++ source code documentation generator)[^1] understand that the comment is a docstring.

It would also be great if you could move the docstring to the place in the cosmology.hpp where we actually declare this function.

[^1]: The generated documentation is hosted on our website. A list of documented C++ classes can be found [here](https://cholla.readthedocs.io/en/latest/Reference/internal-api-ref/index.html). (I'm actively working on addressing some issues that currently prevent doxygen from rendering documentation for a bunch of classes -- including the `Cosmology` class).

mabruzzo (changes requested):
Hi Diego, thanks so much for making all of those changes!

Let me say again that your proposed changes look fantastic! I've left a few more comments/suggestions. The really important comments start with "**IMPORTANT**". The other comments are *much less important*, so feel free to just let me know on each of those if you don't really feel like addressing them (it's also totally fine if you don't want to address any of them).

I'm very appreciative of all the changes you made.

I don't want to overwhelm you, but I did want to suggest one final potential improvement: move this new logic into a separate class. Suppose the the class is called `TabulatedDynamicalDEEOS`. The `Cosmology` would then track a pointer to this class and call the methods of that class. The benefits of doing that are: it makes the logic easy to consider in isolation from the rest of the `Cosmology` class and it can be unit-tested. I'm happy to go into more detail (and help you do this) if you are interested. I also totally understand if you don't want to do that (I really appreciate all the changes you've already made, and making this change definitely isn't necessary -- the code is already great!)

src/cosmology/cosmology.h
astrodiegog: Yes, I forgot to delete this function from testing

src/cosmology/cosmology.h
astrodiegog: This is very helpful! Reading about copy/move constructors and assignment operations, I believe using `std::vector<float>` is the best way to move forward and avoid memory leaks.

src/cosmology/cosmology_functions.cpp
astrodiegog: Yes, I created a separate function for this!

src/cosmology/cosmology_functions.cpp
astrodiegog: Made sure to include this exclamation point, thank you for highlighting this!

astrodiegog:
pre-commit.ci autofix

astrodiegog:
> The `Cosmology` would then track a pointer to this class and call the methods of that class.

I am unsure how best to allocate the memory for this pointer, so I've just saved the method as an object instead of a pointer to this object.

I believe I addressed other comments related to adding some documentation, creating a separate class for the dynamical dark energy implementation, and using `std::vector<float>` instead of just pointers to arrays. 

I am happy to take further comments when people have time!

src/cosmology/dynamical_dark_energy.h
```diff
@@ -0,0 +1,22 @@
+#ifdef COSMOLOGY
```
mabruzzo:
```
/*! \file
 * This file defines \ref TabulatedDynamicalDarkEnergyEoS
 */
```

I would add this snippet to the top of this file so that that doxygen knows to look for documentation in this file

mabruzzo: It would be great if you could get rid of this `#ifdef COSMOLOGY` conditional compilation directive. This totally isn't necessary here, and I'm generally on a mission to remove these as much as possible. (plus, it will help doxygen generate documentation for stuff in this file)

If not, I'm happy to add a PR that does this

src/cosmology/dynamical_dark_energy.h
```diff
@@ -0,0 +1,22 @@
...
+  #include <cmath>
```
mabruzzo: **IMPORTANT:** the include directives should be adjusted so that this file includes headers declaring all of the constructs that it directly uses.

```suggestion
  #include "../global/global.h"
  #include <vector>
```

(clang-tidy may reorder these)

docs/sphinx/Reference/param/Cosmology.md
```diff
@@ -0,0 +1,89 @@
...
+
+---
+
+:::{par:parameter} wDE_file
+
+:Summary: Path to data file describing a redshift-dependent dark energy equation of state
+:Type: {par:typefmt}`str`
+:Default: *None*
+
+We expect increasing redshift values in this file. Each row must have a `z wDE(z)` value with a space delimiter. Table must have an entry at redshift {math}`z=0`. When specified, overrides {par:param}`w0` and {par:param}`wa`.
```
mabruzzo: it might be useful to include a snippet of what this looks like. Maybe the first 3 lines of the table you are using (just to give us a sense of what it looks like)

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,136 @@
+#ifdef COSMOLOGY
+
+  #include <fstream>
+  #include <iostream>
+  #include <vector>
+
+  #include "../io/io.h"
```
mabruzzo: **IMPORTANT:** you should include the header file declaring functions/classes in the source file where you are implement the header/source function.

```suggestion
  #include "../io/io.h"
  #include "dynamical_dark_energy.h"
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,136 @@
+#ifdef COSMOLOGY
```
mabruzzo: You may want to add this snippet so that doxygen knows that it should read this file
```c++
/*! \file
 * This file implements \ref TabulatedDynamicalDarkEnergyEoS
 */
```

mabruzzo: Again, it would be great if you could get rid of this `#ifdef COSMOLOGY` conditional compilation directive. This totally isn't necessary here.

(But no pressure)

src/cosmology/tabulated_dynamicalDE_EoS.h
```diff
@@ -0,0 +1,22 @@
...
+  std::vector<float> dynamicalDE_table_z;
+  std::vector<float> dynamicalDE_table_w;
+  // dynamical dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  std::vector<float> dynamicalDE_table_density;
+
+  /*! Load redshift and dark energy equation of state table z, wDE(z), only called once to setup dynamical DE case */
+  void Setup_DynamicalDE_EquationOfState_(struct Parameters *P);
+
+  /*! Calculate dark energy density normalized to z=0, populate dynamicalDE_table_density */
+  void Set_DynamicalDE_Density();
```
mabruzzo, line 24: (THIS TOTALLY ISN'T NECESSARY), but since you've gone to the trouble of shifting this logic into a nice class, it might be nice to add a constructor:

```suggestion
 private:
  /*! Load redshift and dark energy equation of state table z, wDE(z), only called once to setup dynamical DE case */
  void Setup_DynamicalDE_EquationOfState_(struct Parameters *P);

  /*! Calculate dark energy density normalized to z=0, populate dynamicalDE_table_density */
  void Set_DynamicalDE_Density();
  
 public:
  explicit TabulatedDynamicalDarkEnergyEoS(Parameters *P)
  {
    Setup_DynamicalDE_EquationOfState_(P);
    Set_DynamicalDE_Density();
  }
```

In case you didn't know:
- we can pass `Parameters *P` rather than `struct Parameters *P` (the latter is only necessary in pure C)
- it's generally good practice to annotate a constructor that accepts a single argument with the `explicit` keyword. If you don't do this C++ is allowed to perform implicit casts (while this sounds convenient, it is actually a bad thing because it can lead to very unexpected behavior)

src/cosmology/cosmology.h
```diff
@@ -23,6 +24,15 @@ class Cosmology
   Real w0;
   Real wa;
 
+  /*! Calculate rho_DE(z) / rho_DE(z=0) at some scale factor */
+  Real Get_DE_Density_from_a(Real a);
+
+  /*! Declare dynamical DE equation of state table class */
```
mabruzzo:
```suggestion
  /*! Stores dynamical DE equation of state table */
```

src/cosmology/tabulated_dynamicalDE_EoS.h
mabruzzo: If you decide to make this change, then you would need to replace     

```
tab_dynamicalDE_EoS.Setup_DynamicalDE_EquationOfState_(P);
tab_dynamicalDE_EoS.Set_DynamicalDE_Density();
```

within `cosmology.cpp` with `tab_dynamicalDE_EoS = TabulatedDynamicalDarkEnergyEoS(P);`

mabruzzo (changes requested):
Hi Diego,

This looks exceptionally! I have left a few comments. You can totally ignore all the ones that aren't marked IMPORTANT (the IMPORTANT comments only ask for very minor changes)

EDIT: actually, I meant to ask why you have `dynamical_dark_energy.h` and `tabulated_dynamicalDE_EoS.cpp` (rather than `dynamical_dark_energy.h` and `dynamical_dark_energy.cpp` OR `tabulated_dynamicalDE_EoS.h`  and `tabulated_dynamicalDE_EoS.cpp`). Was that intentional?

mabruzzo:
> > The `Cosmology` would then track a pointer to this class and call the methods of that class.
> 
> I am unsure how best to allocate the memory for this pointer, so I've just saved the method as an object instead of a pointer to this object.
> 
> I believe I addressed other comments related to adding some documentation, creating a separate class for the dynamical dark energy implementation, and using `std::vector<float>` instead of just pointers to arrays.
> 
> I am happy to take further comments when people have time!

Just in case your curious, the way I would handle the pointer allocation stuff is I would make a unique pointer. This would involve:
- including `<memory>` at the top of `comology.h`
- replacing `TabulatedDynamicalDarkEnergyEoS tab_dynamicalDE_EoS`; in `cosmology.h` with `std::unique_ptr<TabulatedDynamicalDarkEnergyEoS> tab_dynamicalDE_EoS`;
- replacing
  ```c++
    tab_dynamicalDE_EoS.Setup_DynamicalDE_EquationOfState_(P);
    tab_dynamicalDE_EoS.Set_DynamicalDE_Density();
    ```
    with `tab_dynamicalDE_EoS = std::unique_ptr<TabulatedDynamicalDarkEnergyEoS>( new TabulatedDynamicalDarkEnergyEoS(P));`
  (this assumes that you added a constructor like I mentioned int the comments)

You would be able to check if the underlying pointer is empty by calling `tab_dynamicalDE_EoS.get() == nullptr`. And you could call methods on this pointer by writing `tab_dynamicalDE_EoS->Get_DynamicalDE_Density_from_a(a);`

(To be clear, you definitely don't have to do this -- I just wanted to explain how it would be accomplished)

src/cosmology/dynamical_dark_energy.h
astrodiegog: removed !

astrodiegog:
> Was that intentional?

I was flipping between both and forgot to decide ! I have changed `dynamical_dark_energy.h` to `tabulated_dynamicalDE_EoS.h`

src/cosmology/tabulated_dynamicalDE_EoS.h
astrodiegog: Having a constructor makes the code look much cleaner, but I am having trouble integrating this. When I compile, I receive the following errors:

```
src/cosmology/cosmology.cpp: In constructor ‘Cosmology::Cosmology()’:
src/cosmology/cosmology.cpp:7:26: error: no matching function for call to ‘TabulatedDynamicalDarkEnergyEoS::TabulatedDynamicalDarkEnergyEoS()’
    7 | Cosmology::Cosmology(void) {}
      |                          ^
In file included from src/cosmology/../cosmology/cosmology.h:13,
                 from src/cosmology/cosmology.cpp:3:
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:22:12: note: candidate: ‘TabulatedDynamicalDarkEnergyEoS::TabulatedDynamicalDarkEnergyEoS(Parameters*)’
   22 |   explicit TabulatedDynamicalDarkEnergyEoS(struct Parameters *P);
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:22:12: note:   candidate expects 1 argument, 0 provided
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:10:7: note: candidate: ‘TabulatedDynamicalDarkEnergyEoS::TabulatedDynamicalDarkEnergyEoS(const TabulatedDynamicalDarkEnergyEoS&)’
   10 | class TabulatedDynamicalDarkEnergyEoS
      |       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:10:7: note:   candidate expects 1 argument, 0 provided
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:10:7: note: candidate: ‘TabulatedDynamicalDarkEnergyEoS::TabulatedDynamicalDarkEnergyEoS(TabulatedDynamicalDarkEnergyEoS&&)’
src/cosmology/../cosmology/tabulated_dynamicalDE_EoS.h:10:7: note:   candidate expects 1 argument, 0 provided
```

I think this has to do with the definition `TabulatedDynamicalDarkEnergyEoS tab_dynamicalDE_EoS;` in the `Cosmology` class being created without using its constructor. From what I found online, people seem to recommend storing a pointer instead of the object itself. 

This would also tie in with your most recent comment about using `std::unique_ptr` to store a pointer to this table !

astrodiegog:
pre-commit.ci autofix

astrodiegog:
Would like to work on storing `TabulatedDynamicalDarkEnergyEoS` as pointer in `Cosmology` in next couple days!

src/cosmology/tabulated_dynamicalDE_EoS.h
mabruzzo: You are exactly right! This pertains to default constructors.

I can provide you with a little more context:
- the error is pointing to the definition of `Cosmology`'s default constructor. Since `Cosmology`'s default constructor doesn't invoke any particular constructor for its `tab_dynamicalDE_EoS` data member,[^1] the compiler will try to invoke that data member's default constructor.
- in the code's current state (without the change I suggested here), the fact that you haven't defined any constructor for `TabulatedDynamicalDarkEnergyEoS` means that the compiler will implicitly define a default constructor that simply calls the default constructor of each data member in `TabulatedDynamicalDarkEnergyEoS`
- when you implement the change I suggested here, the fact that you are implementing any constructor for  `TabulatedDynamicalDarkEnergyEoS` means that the compiler stops implementing a default constructor `TabulatedDynamicalDarkEnergyEoS`

There are several ways to resolve this issues
1. Simple approach: just add a default-constructor to `TabulatedDynamicalDarkEnergyEoS`. This might look like:
   ```c++
    public:
     TabulatedDynamicalDarkEnergyEoS() = default;

     explicit TabulatedDynamicalDarkEnergyEoS(Parameters *P)
     {
       Setup_DynamicalDE_EquationOfState_(P);
       Set_DynamicalDE_Density();
     }
     ```
     There are multiple other ways to do this.[^2]

 2. As you already described, you can convert `Cosmology::tab_dynamicalDE_EoS` so that it's a `std::unique_ptr<TabulatedDynamicalDarkEnergyEoS>`.
    - This is what I would recommend.
    - If you have trouble using `std::unique_ptr`, it may be easier to first write a commit that uses regular pointers (i.e. a `TabulatedDynamicalDarkEnergyEoS*`) and then write a subsequent commit with `std::unique_ptr`. (Also feel free to ask me for help).
    - The reason this works is that `std::unique_ptr` has a default constructor -- by default it stores a null pointer.[^3]
 
3. You can convert `Cosmology::tab_dynamicalDE_EoS` so that it's a `std::optional<TabulatedDynamicalDarkEnergyEoS>`.
   - this theoretically provides certain performance advantages over using a `std::unique_ptr`, but they aren't significant in the current context.[^4]
   - since you are still familiarizing yourself with C++, it probably makes more sense for you to use a `std::unique_ptr`

4. You can convert `Cosmology::tab_dynamicalDE_EoS` so that it's a container (e.g. a std::vector) of `TabulatedDynamicalDarkEnergyEoS` instances. **This doesn't make any real sense in the current context.**
 
Personally, I generally view solutions like options 2 and 3 as more elegant than option 1:
- If you have a default constructor (option 1), its possible to have an instance of `TabulatedDynamicalDarkEnergyEoS` that is empty. In this case, its the programmer's has to know whether the table is empty or not to decide whether they can call the `Get_DynamicalDE_Density_from_a` method.
- If you don't have a default constructor (options 2 or 3) then it's impossible to have an empty table
- practically speaker, the distinction isn't very significant since the class is so simple
 
 [^1]: **ASIDE** when defining a class's constructor, you can specify which of the constructors to call for creating a data member using [member initializer lists](https://www.en.cppreference.com/w/cpp/language/initializer_list.html).
[^2]: The `= default;` syntax that I used is a way to explicitly tell the compiler to define the same default constructor that it had been implicitly defining before this change.
[^3]: If you use a regular pointer, it doesn't have any constructor. It will be initialized in an indeterminate state (just like in C).
[^4]: For some context, a `std::unique_ptr` or `std::optional` can both be used to signal that it wraps a value that may or may not exist. If the value does exist, `std::unique_ptr` holds a pointer to the memory containing that value whereas a `std::optional` directly holds the memory containing the value. Thus, using a `std::optional` can be faster since there are fewer memory dereferences. In the way that `Cosmology::tab_dynamicalDE_EoS` is currently used (and the overhead of branch-prediction), I doubt this performance differential really matters.

mabruzzo:
@astrodiegog, Now that you have converted `TabulatedDynamicalDarkEnergyEoS` into its own class, it would be great if we could write a simple unit test.

You've already gone above and beyond in this PR, and learning how to use GoogleTest can be a little involved. So if you give me a sample file that this class reads, I would be happy to do that myself (I can open a new PR that would be reviewed after this one is merged)

astrodiegog:
pre-commit.ci autofix

src/cosmology/tabulated_dynamicalDE_EoS.h
astrodiegog: Gotcha ! This makes sense to me. Thank you for taking the time to provide some detailed context, and the different ways to go about solving this issue.

I opted for option 2 (using `std::unique_ptr`)! I just pushed these changes

astrodiegog:
> You've already gone above and beyond in this PR, and learning how to use GoogleTest can be a little involved. So if you give me a sample file that this class reads, I would be happy to do that myself (I can open a new PR that would be reviewed after this one is merged)

Thanks ! I just sent a sample file via Slack

evaneschneider:
pre-commit.ci autofix

src/cosmology/tabulated_dynamicalDE_EoS.h
```diff
@@ -0,0 +1,41 @@
...
+{
+ private:
+  std::vector<float> dynamicalDE_table_z;
+  std::vector<float> dynamicalDE_table_w;
+
+  // dynamical dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  std::vector<float> dynamicalDE_table_density;
+
+  /*! Load redshift and dark energy equation of state table z, wDE(z), only called once to setup dynamical DE case */
+  void Setup_DynamicalDE_EquationOfState_(struct Parameters *P);
```
mabruzzo:
```suggestion
  void Setup_DynamicalDE_EquationOfState_(const std::string& path);
```

src/cosmology/tabulated_dynamicalDE_EoS.h
```diff
@@ -0,0 +1,41 @@
...
+
+  // dynamical dark energy energy density normalized to the present-date rho_DE(z) / rho_DE(z=0)
+  // std::vector<float> dynamicalDE_table_density;
+
+  /*! Interpolate dynamicalDE_table_density to find rhoDE(z) / rhoDE(z=0) at z=1/a - 1 */
+  Real Get_DynamicalDE_Density_from_a(Real a);
+
+  explicit TabulatedDynamicalDarkEnergyEoS(Parameters *P)
+  {
+    Setup_DynamicalDE_EquationOfState_(P);
```
mabruzzo:
```suggestion
  explicit TabulatedDynamicalDarkEnergyEoS(const std::string& path)
  {
    Setup_DynamicalDE_EquationOfState_(path);
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,138 @@
...
+    dynamicalDE_table_density.push_back(onez_3 * exp(3. * cumulative_integral));
+  }
+}
+
+void TabulatedDynamicalDarkEnergyEoS::Setup_DynamicalDE_EquationOfState_(struct Parameters *P)
+{
+  if (P->wDE_file.empty()) {
+    chprintf("wDE_file not found in parameter file \n");
+    exit(1);
+  }
```
mabruzzo:
```suggestion
void TabulatedDynamicalDarkEnergyEoS::Setup_DynamicalDE_EquationOfState_(const std::string& path)
{
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+  chprintf("Loading wDE info... \n");
+
+  std::fstream in(P->wDE_file);
```
mabruzzo:
```suggestion
  std::fstream in(path);
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,138 @@
...
+      v.push_back(std::vector<float>());
+
+      while (ss >> value) {
+        v[i].push_back(value);
+      }
+      i += 1;
+    }
+    in.close();
+  } else {
+    chprintf(" Error: Unable to open DE equation of state file: %s\n", P->wDE_file);
```
mabruzzo:
```suggestion
    chprintf(" Error: Unable to open DE equation of state file: %s\n", path.c_str());
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,138 @@
...
+    dynamicalDE_table_z.push_back(v[i][0]);
+    dynamicalDE_table_w.push_back(v[i][1]);
+  }
+
+  for (i = 0; i < n_lines - 1; i++) {
+    if (dynamicalDE_table_z[i] > dynamicalDE_table_z[i + 1]) {
+      chprintf(
+          " ERROR: equation of state must be ordered such that redshift is increasing "
+          "as the rows increase in the file\n",
+          P->wDE_file);
```
mabruzzo:
```suggestion
          path.c_str());
```

mabruzzo (approved):
I took a look at the failing tests, and I realized that the solution is relatively simple: stop passing the `Parameters` struct into the constructor of `TabulatedDynamicalDarkEnergyEos`. Instead you can directly pass in the path to the data file. (This also has benefits for testing -- e.g. see PR #506)

I have attached suggested changes on how to easily accomplish this.

Once you resolve the test failure, everything looks great to me! (Thanks again for making all of the changes)

evaneschneider:
pre-commit.ci autofix

src/cosmology/cosmology.h
```diff
@@ -23,6 +25,15 @@ class Cosmology
...
   Real wa;
 
+  /*! Calculate rho_DE(z) / rho_DE(z=0) at some scale factor */
+  Real Get_DE_Density_from_a(Real a);
+
+  /*! Stores pointer to dynamical DE equation of state table class */
+  std::unique_ptr<TabulatedDynamicalDarkEnergyEoS> tab_dynamicalDE_EoS;
+
+  /*! Indicate whether the simulation is configured to use a DynamicalDE EOS table */
+  bool Using_DynamicalDE_Table() { return tab_dynamicalDE_EoS.get() != nullptr; }
```
mabruzzo:
```suggestion
  bool Using_DynamicalDE_Table() { return tab_dynamicalDE_EoS != nullptr; }
```

src/cosmology/cosmology.cpp
```diff
@@ -21,6 +21,12 @@ void Cosmology::Initialize(struct Parameters *P, Grav3D &Grav, Particles3D &Part
   w0      = P->w0;
   wa      = P->wa;
 
+  if (P->wDE_file.length() != 0) {
+    // set pointer to DynamicalDE EOS
+    tab_dynamicalDE_EoS =
+        std::unique_ptr<TabulatedDynamicalDarkEnergyEoS>(new TabulatedDynamicalDarkEnergyEoS(P->wDE_file));
```
mabruzzo:
```suggestion
    tab_dynamicalDE_EoS = std::make_unique<TabulatedDynamicalDarkEnergyEoS>(P->wDE_file);
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+  chprintf("Loading wDE info... \n");
+
+  std::fstream in(path);
+  std::string line;
+  std::vector<std::vector<float>> v;
+  int i = 0;
+  if (in.is_open()) {
+    while (std::getline(in, line)) {
+      if (line.find("#") == 0) continue;
```
mabruzzo:
```suggestion
      if (line.find('#') == 0) continue;
```

src/cosmology/tabulated_dynamicalDE_EoS.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+      float value;
+      std::stringstream ss(line);
+      v.push_back(std::vector<float>());
```
mabruzzo:
```suggestion
      v.emplace_back();
```

mabruzzo (approved):
The tests are failing right now because the linter wants you to make a few changes.

To try to help out, I've suggested changes the required changes (I think I caught all of them).
- Each changed line accomplishes the exact same thing as the old version of the code. The new version is either slightly more performant or slightly more direct
- if you have any questions about the suggested changes just let me know

astrodiegog:
pre-commit.ci autofix

PR #502, opened 2026-04-01
Rm `COOLING_GPU` & `CLOUDY_COOL` macros, and the cooling-build type

mabruzzo:
To be reviewed after #485 has been merged

----------

This PR finally removes the `COOLING_GPU` & `CLOUDY_COOL` macros (they haven't been necessary for quite a while) and the cooling-build type (as well as the cloudy-build type). I have migrated the system tests so that they are run as part of the hydro test suite.


docs/sphinx/Physics/CoolingChemistry.md
```diff
@@ -26,34 +26,22 @@ Supported options include:
...
 
 * - Name
   - Meaning
-  - Macros that make this the default
 * - "none"
-  - No chemistry or cooling
-  - N/A
+  - No chemistry or cooling (this is the default)
 * - "tabulated-cloudy"
   - path to data file used by "tabulated-cooling" solver
```
evaneschneider: I think this line is incorrect (this should be a description like that given in the paragraph above).

docs/sphinx/Physics/CoolingChemistry.md
evaneschneider:
```suggestion
This solver uses the same analytic fit as "piecewise-cie" above 1e4K and extends the fit at lower temperatures using the analytic fit from [Koyama & Inutsuka (2002)](https://ui.adsabs.harvard.edu/abs/2002ApJ...564L..97K/abstract), eq. 4 & 5.
```

evaneschneider:
```suggestion
Cooling and heating are computed from the table using bilinear interpolation.
```

evaneschneider:
```suggestion
Several of the solvers described above can be configured so that they also model the effect of photoelectric heating.
```

evaneschneider:
```suggestion
When photoelectric heating is enabled, you can control the assumed average number density (used to scale the normalization of the assumed ISRF) with the {par:param}`chemistry.photoelectric_n_av_cgs` parameter.
```

evaneschneider:
```suggestion
- `"chemistry-gpu"`: uses logic implemented directly as part of Cholla to run chemistry on GPUs (cooling functions match those in the Grackle library)
```

docs/sphinx/Reference/param/Chemistry.md
evaneschneider: Should this line be modified now (or deleted)?

evaneschneider (approved):
A few minor suggestions in the docs, otherwise this looks good!

PR #503, opened 2026-04-07
[Param Update 1 of 2] Quoted string parameters in parameter files

mabruzzo:
This should not be reviewed until after #535 has been merged

---------

This PR modifies Cholla so that all strings in the parameter files are quoted.

> [!important]
> This would be a backwards incompatible change. As such, I realize that it may take a long time to get merged (or it may be closed).

## Motivation
If this PR is merged, Cholla's parameter files will be a strict subset of the TOML format. This has a few benefits:

1. Handling of whitespace in string values is no longer ambiguous.

2. We can start using an external TOML parsing library (there are a handful of nice C++ implementations that we can use simply by vendoring a single header file). This lets us get rid of the underlying parsing logic, and lets us fully use some of TOML's nice features (namely tables and arrays).

3. It will be possible to easily read the input parameter files with any valid TOML parsing library. This could be useful in analysis.

4. Relatedly, we could more easily programmatically modify Cholla input files (yeah, this is niche, but when it comes up, it's quite nice!)

## What has changed

- This PR has made minimal internal changes to make this all work. I do think we should switch to an external TOML library, but I wanted to keep the diff minimal to make this PR.
- It also has tweaked the system-test logic to properly forward arguments
- It also introduces a python script that can be used to automatically update the legacy input files.
- Additionally, it includes updates to the documentation describing parameter files


docs/sphinx/WritingParamFiles.md
```diff
@@ -40,11 +39,54 @@ Currently, Cholla does **NOT** properly handle whitespace around the equal sign
...
+Currently, we require strings to be composed of ASCII characters.
+
+The following snippet illustrates the 2 supported approaches for specifying a string.
+
+```toml
+indir="path/to/input-data"
+outdir='path/to/output'
+```
+
+In the snipet, the value associated with `indir` is a "basic string" (i.e. it is enclosed by a pair of `"` characters), while the value associated with `outdir` is a "raw string" (it is enclosed by a pair of `'` characters.
```
evaneschneider:
```suggestion
In the snippet, the value associated with `indir` is a "basic string" (i.e. it is enclosed by a pair of `"` characters), while the value associated with `outdir` is a "raw string" (it is enclosed by a pair of `'` characters.
```

src/io/ParameterMap.cpp
```diff
@@ -88,9 +92,19 @@ param_details::TypeErr param_details::try_double_(const std::string& str, double
 
 param_details::TypeErr param_details::try_string_(const std::string& str, std::string& val)
 {
-  // mostly just exists for consistency (every parameter can be considered a string)
-  // note: we may want to consider removing surrounding quotation marks in the future
-  val = str;  // we make a copy for the sake of consistency
+  // printf("trying to parse: %s\n", str.c_str());
+  std::pair<std::size_t, std::string> rslt = io::try_parse_param_str(str);
+  // printf("-> pos = %d, (actual size = %d)\n", (int)rslt.first, (int)str.size());
```
evaneschneider, line 97: Do we need these commented-out printfs? They look like maybe they're leftover from debugging.

src/system_tests/system_tester.cpp
```diff
@@ -358,10 +358,26 @@ void system_test::SystemTestRunner::launchCholla()
 {
   // Launch Cholla. Note that this dumps all console output to the console
   // log file as requested by the user.
+  //
+  // As we've highlighted in the implementation of `ParamArgListBuilder::param` (for
+  // strings), the choice to launch Cholla by calling the system function is quite
+  // messy since the argument to `system` is executed by the shell (this leads to
+  // messy) argument quoting. A more robust is to bypass the shell.
```
mageejake:
```suggestion
  // messy) argument quoting. A more robust method is to bypass the shell.
```

tools/convert-unquoted-params.py
```diff
@@ -0,0 +1,222 @@
+#!/usr/bin/env python3
+"""
+Convert a legacy parameter file so the string-parameters are now quoted.
+Importantly, the program assumes that the file format is valid.
+"""
+
+# for portability, only use built-in package packages present in python 3.7
```
evaneschneider:
```suggestion
# for portability, only use built-in packages present in python 3.7
```

evaneschneider:
Looks good to me once the parameter files are updated for the system tests.

mageejake (commented):
Left a small grammatical comment on system_tester.cpp, otherwise everything (that I could understand) looks good to me!

evaneschneider (approved):

PR #507, opened 2026-04-21
WIP: Move Parsing of parameters governing the "Main-Loop"

mabruzzo:
EDIT: this needs to be adjusted now that the cosmo-ics PR is merged

This PR moves parsing of the parameters that govern the "main-loop" out of `global.(cpp|h)` and do it directly in the main function.

As I note in the function, it might be a little better to aggregate the variables inside a tiny struct, but I wanted to gauge feelings before attempting that change


src/main.cpp
```diff
@@ -89,13 +89,23 @@ int main(int argc, char *argv[])
...
 
+  // read in a few parameters only used in the top-level main function (maybe aggregate in a struct?)
+  const double tout = pmap.value<double>("tout");  // aborts if missing
+  CHOLLA_ASSERT(parms->tout >= 0.0, "tout parameter must be non-negative");
+  const double outstep    = pmap.value<double>("outstep");  // aborts if missing
+  const int output_always = pmap.value_or("output_always", 0);
+  CHOLLA_ASSERT((output_always == 0) or (output_always == 1), "output_always must be 1 or 0.");
+  const int n_steps_limit = pmap.value_or("n_steps_limit", -1);  // negative values indicate that there is no limit
+  // the following is unused (can we remove it?)
+  // const int n_steps_output = pmap.value_or("n_steps_output", 0);
```
mabruzzo: Can we remove this entirely?

src/cosmology/io_cosmology.cpp
```diff
@@ -53,7 +53,6 @@ void Cosmology::Load_Scale_Outputs(struct Parameters *P)
 void Cosmology::Set_Scale_Outputs(struct Parameters *P)
 {
   if (P->scale_outputs_file[0] == '\0') {
-    chprintf(" Output every %d timesteps.\n", P->n_steps_output);
```
mabruzzo, line 56: It's ok to delete the line. It turns out that this statement was effectively a "lie" (the `P->n_steps_output` value was never actually used for anything, at all).

I imagine that this was never deleted when we stopped using `P->n_steps_output`

PR #511, opened 2026-04-28
Standardize Cool-Components and Cool Recipies (Cool Refactor 3/3)

mabruzzo:
To be reviewed after #510 is merged. This is the third in a sequence of 3 PRs to refactor and standardize cooling recipes and components.

------------

This finishes refactoring cooling recipe -- they are now implemented in a more consistent manner.

**TODO**: Add more context


hannahjleary (approved):

mabruzzo:
That was an annoying bug to fix. It turns out that while `nvcc` treats `__forceinline__` as a keyword, `hipcc` defines it as a macro in its header files.

I feel like I've solved this exact problem in the past

PR #521, opened 2026-06-08
Non-Periodic Potential Boundary Refactor

mabruzzo:
To be reviewed after PR #535 is merged

------

This PR lightly refactors the way in which we set the potential at the boundaries. This PR shouldn't actually change any logic. Instead, it's equivalent to hoisting the if-statement for switching between types of models for setting the potential boundary condition out of the nested for-loop that iterates over the ghost cells.

This should definitely be faster than what we were previously doing (since there's less branching).

This was spurred on by the upcoming PR where I change the way that we track model objects. This has been a long time coming and there is more that I want to do, but I'm leaving that for a followup PR.


mabruzzo:
As part of this PR, I tried to run a basic test to check if this worked. Essentially, I compiled and ran a test-problem of a disk (without feedback) that evolved the simulation over a single timestep using the dev-branch and I did the same thing using another commit and checked for equality of the potential array from the resulting pair of snapshots.

>[!important]
>The short answer is: yes, the potential in the boundaries (i.e. what this PR changes) hasn't changed. Thus this PR is ready to go.

However, my expectation that the we could perform this check by using the `h5diff` command line tool to check for bitwise equality between arrays was wrong. This caused me to spend a **LOT** of time on this, and I want to record what happened for down below for posterity. The moral of this story is that our current algorithm for particle deposition isn't strictly deterministic. It uses atomic operations to accumulate the density and since the floating point addition isn't associative/commutative, the results aren't bitwise deterministic.

------
While running this check, I made a point of removing `-Ofast` from the compiler flags and I made sure to only run the calculation with a single block grid (these could conceivably change the results). I thought that this would guarantee that equivalent versions of the codebase would produce bitwise-identical results (spoiler it doesn't).

So, when I ran this check, I found that the computed potential in the outputs of the first cycle was not bitwise-identical. Then I decided to manually bisect to the commit where I introduced the error. This was somewhat time-consuming, and to my surprise I found that all of the commits in this branch produced different results from the dev branch.

At this point,  I finally decided to start reading the output files into python and start plotting the differences in the potential. After digging out and fixing up some code that I previously employed for this kind of thing, I found the following:
- there were no differences in the potential (meaning that this PR can be merged)
- however I found relative errors in the potential with magnitudes of ~1e-16.
- curiously, these errors were randomly scattered (but there was a higher density of errors in the midplane)

The following collapsible section show what these errors look like. Note that these plots include the ghost zones (the black dashed lines show where the active zone starts).

<details>

<summary>The projected potential and projected errors in the potential</summary>

potential-proj-diff

</details>

I found this all to be somewhat worrisome (maybe it's a sign of a race condition or undefined behavior?). Anyway, after comparing some more data, I realized that the difference arises during particle deposition. **It turns out that we use atomic operations during particle deposition** (with hindsight, this should have been obvious -- I just totally forgot about it). The following collapsible section illustrates the deposited particle density and the difference between runs for the dev branch and the commit at the head of this branch.

If the number of particles was very low, this wouldn't have been an issue, but there were 172,124 particles within the disk in these simulations.

<details>

<summary>The projected deposited density and projected errors in the deposited density</summary>

deposited-density-diff

</details>

mabruzzo:
I made the most recent "force-push" after I rebased the branch onto the most recent version of dev.

NOTE: I did all of the testing described in the prior comment before rebasing and before making the most recent commit (tweaking 2 lines to address a clang-tidy warning). But, I'm extremely confident that this won't break things.

hannahjleary (approved):
Looks great!

PR #526, opened 2026-07-08
Adjustments to make CI catch all clang-tidy errors

mabruzzo:
It has come to my attention that clang-tidy errors in .cpp files tend to slip through CI.

This is because the Makefile would only report the errors of the second execution of clang-tidy that would exit and it appears that the clang-tidy almost always runs more quickly for the .cpp files (perhaps the .cu files are simply more numerous).

The first commit tweaks the Makefile's `tidy` recipe so that we now report the correct result. This causes CI to correctly report a linting error (rather than suppress it). The subsequent commit commit tries to address all of the warnings that piled up

---

ASIDE: if we make a build-directory, there are more robust ways to run clang-tidy that would let us distribute clang-tidy runs across an arbitrary number of CPU cores, in the same way we can distribute regular compilation commands. These approaches involve the creation of a compile_commands.json file -- but that is an idea for a subsequent PR

### DIFF
--- a/src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
+++ b/src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
@@ -134,6 +134,7 @@ TEST(tALLTabulatedDynamicalDarkEnergyEoSDeathTest, ThreeElemsOnLine2)
 
 // define parametrized tests where we try to read files without real contents
 
+// NOLINTNEXTLINE(readability-identifier-naming)
 class tALLTabulatedDynamicalDarkEnergyEoSNoContentsDeathTest : public testing::TestWithParam<FileVariation>
 {
 };
--- a/src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
+++ b/src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
@@ -46,6 +46,7 @@ const std::string GOOD_CONTENTS_ = R"LITERAL(# z, w
 # this is a random meaningless comment!
 1.392540755881421788e-02 -9.772263520514015145e-01)LITERAL";
 
+// NOLINTNEXTLINE(readability-identifier-naming)
 class tALLTabulatedDynamicalDarkEnergyEoS : public testing::TestWithParam<FileVariation>
 {
 };
--- a/src/io/io.cpp
+++ b/src/io/io.cpp
@@ -227,7 +227,7 @@ H5Space1D &H5Space1D::operator=(H5Space1D &&other) noexcept
 H5Space1D &H5Space1D::ensure_dim(hsize_t dim)
 {
   if (this->id_ == H5I_INVALID_HID) {  // <- first time setting dim
-    this->dim_ = std::unique_ptr<hsize_t>(new hsize_t{dim});
+    this->dim_ = std::make_unique<hsize_t>(dim);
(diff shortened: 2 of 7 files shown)


mabruzzo:
This entire chain of commit feels gives off vibes similar to [*If you Give a Mouse a Cookie*](https://en.wikipedia.org/wiki/If_You_Give_a_Mouse_a_Cookie). The change to needed to address one warning clang-tidy warning, makes a new warning... (I do think its a net positive, I just wish it was a little easier/quicker to run the linter)

src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
```diff
@@ -46,6 +46,7 @@ const std::string GOOD_CONTENTS_ = R"LITERAL(# z, w
 # this is a random meaningless comment!
 1.392540755881421788e-02 -9.772263520514015145e-01)LITERAL";
 
+// NOLINTNEXTLINE(readability-identifier-naming)
 class tALLTabulatedDynamicalDarkEnergyEoS : public testing::TestWithParam<FileVariation>
```
mabruzzo, line 50: This is necessary since our googletest naming convention doesn't adhere to the class naming conventions.

src/cosmology/tabulated_dynamicalDE_EoS_tests.cpp
```diff
@@ -133,6 +134,7 @@ TEST(tALLTabulatedDynamicalDarkEnergyEoSDeathTest, ThreeElemsOnLine2)
 
 // define parametrized tests where we try to read files without real contents
 
+// NOLINTNEXTLINE(readability-identifier-naming)
 class tALLTabulatedDynamicalDarkEnergyEoSNoContentsDeathTest : public testing::TestWithParam<FileVariation>
```
mabruzzo, line 138: Again, this is necessary since our googletest naming convention doesn't adhere to the class naming conventions.

src/io/io.cpp
```diff
@@ -884,17 +884,22 @@ void Grid3D::Print_Grid_Stats(void)
...
-  Real *HeII_density  = &host_field_ptr[H.n_cells * field_info.field_id("HeII_density").value()];
-  Real *HeIII_density = &host_field_ptr[H.n_cells * field_info.field_id("HeIII_density").value()];
-  Real *e_density     = &host_field_ptr[H.n_cells * field_info.field_id("e_density").value()];
+  Real *density       = get_ptr_or_abort("density");
+  Real *HI_density    = get_ptr_or_abort("HI_density");
+  Real *HII_density   = get_ptr_or_abort("HII_density");
+  Real *HeI_density   = get_ptr_or_abort("HeI_density");
+  Real *HeII_density  = get_ptr_or_abort("HeII_density");
+  Real *HeIII_density = get_ptr_or_abort("HeIII_density");
+  Real *e_density     = get_ptr_or_abort("e_density");
```
mabruzzo, line 902: These changes were made to make Cholla loudly error out if we try to execute this logic and one of the accessed fields is missing.

Before this commit, an exception would have been raised, just not in the way that clang-tidy likes. Plus, this is now a lot more explicit about the underlying error if an error does arise.

mabruzzo (commented):
I just wanted to leave a few comments to help with reviewing this

evaneschneider (approved):
Looks good.

PR #529, opened 2026-07-15
Add uint8_t static cast to avoid hicpp-signe-bitwise

### LINKED ISSUE #528: FrozenKeyIdxBiMap.h contains a signed integer operand with a binary bitwise operator
Lines 63 and 77 of FrozenKeyIdxBiMap.h read:

`hash = (hash ^ key[i]) * FNV1A_PRIME;`

where key is potentially signed because UTF-8 or extended ASCII characters would be a negative number.  Regardless of whether this would actually happen, it triggers `hicpp-signe-bitwise` during `clang-tidy`.

The solution to this it so cast key to an unsigned 8-bit integer.  Will create a pull request to fix this issue.

brantr:
If key is used as a char or an element of string_view, it can be negative depending on its contents.  To avoid hicpp-signe-bitwise in clang-tidy, it can be recast as a uint8_t.  Closes #528.


mabruzzo (approved):
Looks good to me! Thanks for making this PR!

I'm not sure that this actually supports unicode characters, but that shouldn't come up.[^1]

[^1]: For some context, my concern is about C++ implementations where `char` is equivalent to `signed char` (the standard doesn't specify whether `char` is equivalent to `signed char` or `unsigned char`). It *might* work out (I'm not sure off the top of my head), but the safer thing to do is to use C++20's, [`std::bit_cast`](https://en.cppreference.com/cpp/numeric/bit_cast) rather than a regular cast. This is all moot since we currently use C++17 and unicode probably won't come up.

PR #530, opened 2026-07-24
Update cuda and rocm dockerfiles to include clang-tidy

mabruzzo:
The dockerfiles have been updated to install our preferred version of clang-tidy. The goal of this change is three-fold:

1. It will let me test discrepancies between clang-tidy on different machines.[^1]

2. It will let us set up GitHub Actions to directly run clang-tidy. This will allow people to see linting errors without requiring credentials to log into Jenkins.

3. This paves the way for development in containers. In other words, someone could locally spin up a docker container using one of our images and compile the code there (even if they don't own a GPU). VSCode has builtin machinery for this purpose called [dev containers](https://code.visualstudio.com/docs/devcontainers/create-dev-container). This has has 2 benefits:
   - if there are issues with clang-tidy lints it will be easier to reproduce them locally.[^2]
   - this is probably the best path to unlocking IDE-like features, with tools like the [clangd](https://clangd.llvm.org/) language server.[^3] I've used this with Grackle, and the real-time feedback makes dealing with clang-tidy **much** easier.

I think this is also a useful step towards eventually updating the clang-tidy version (this is desirable if we ever want to use something like the clangd language server).

[^1]: I know that the c++ standard libraries sometimes have different headers on different architectures. Its highly unlikely, but I want to make sure that this isn't the cause of discrepancies.
[^2]: In practice, clang-tidy shouldn't have a strong dependence on its patch version (e.g. 17.0.1 vs 17.0.6 shouldn't make a difference). I'm hoping to sort this out.
[^3]: Some extra work is definitely required before we can get that working.

### DIFF
--- a/.github/workflows/build_image.yml
+++ b/.github/workflows/build_image.yml
@@ -34,6 +34,13 @@ env:
 jobs:
   build-and-push-image:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # we list suffix (of the image name) path (to the dockerfile) pairs
+        sp_pair:
+          - [ROCm, docker/rocm/Dockerfile]
+          - [cuda, docker/cuda/Dockerfile]
+
     # Sets the permissions granted to the `GITHUB_TOKEN` for the actions in
     # this job.
     permissions:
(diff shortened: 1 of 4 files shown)


evaneschneider (approved):
Looks good!

PR #532, opened 2026-08-10
Scalar Flux Bug Fix

mabruzzo:
## Overview

This PR introduces bugfixes for the HLL, HLLC and Roe Riemann Solvers when computing fluxes for passive scalars and using the piecewise constant method for reconstruction. I'm pretty sure that these bug were introduced by PR #377.

This bug affects any simulation that used passive scalars AND:

- HLL, HLLC or Roe Riemann Solvers with the SIMPLE integrator and piecewise constant reconstruction (I'd be stunned if this actually happened)

- HLL, HLLC or Roe Riemann Solvers with the Van Leer integrator (since piecewise constant reconstruction is always used to predict state at the half timestep).

In practice, there were actually 2 kinds of bugs:
1. The HLLC and Roe Riemann Solver both had the same bug that produces undefined behavior when computing the passive scalar flux(es) with piecewise constant reconstruction (the computed flux used uninitialized values).
2. The HLL solver had a separate bug, that simply produced incorrect flux(es) with piecewise constant reconstruction (however, they were always wrong in a well-defined way). I stumbled onto this while addressing the other bug

> [!important]
> To be clear, the bug invoking undefined behavior was present on all systems. It just manifested differently based on the hardware/compiler-toolchain. While we only noticed it on Betty, it was DEFINITELY present on other systems!

### General Background
Prior to PR #377, reconstructed values of the (magneto)hydrodynamic variables were computed for the entirety of the local grid in one kernel-call and they were subsequently used in a separate kernel call to compute fluxes.

Bob sought to change this in PR #377. He wanted to start fusing the logic for reconstruction and flux calculations into a single kernel. He made that precise change for piecewise constant reconstruction. However, the other reconstruction mechanisms are still performed ahead of time (I suspect we have a hybrid solution because he ran out of time to finish it all up).

Unfortunately, he made minor oversights in the HLL, HLLC, and Roe Riemann solvers.

### Bug in HLLC and Roe Riemann Solvers

The problem for the HLLC and Roe Riemann solvers is that a pair of vestigial temporary arrays for holding passive scalar densities (named `dscl` & `dscr` in HLLC or `dscalarl` & `dscalarr` in Roe) persisted in the flux calculations. These arrays are temporarily used for higher-order reconstruction. However, for piecewise piecewise constant reconstruction, these arrays are never initialized. Since the scalar fluxes are computed from values in these arrays, they are have undefined values with piecewise constant reconstruction.

### Bug in the HLL Riemann Solver

This bug pertains to the ``reconstruction::InterfaceState`` class:
- when reconstruction and the riemann solver are part of a fused kernel (reminder: this currently just applies to piecewise constant reconstruction), the ``reconstruction::Reconstruct_Interface_States`` function fills instances of the ``reconstruction::InterfaceState`` are filled that represent the right and left interfaces.
- for higher order reconstruction methods, all of the Riemann Solvers currently fill in the various members of the ``reconstruction::InterfaceState`` class with the corresponding values from reconstructed values that were computed ahead of time.

The problem specifically relates to the treatment of ``reconstruction::InterfaceState::scalar[i]``. The convention observed in  ``reconstruction::Reconstruct_Interface_States`` for piecewise constant reconstruction (as well as in other Riemann Solvers) is that this is a mass fraction. Unfortunately, for higher order reconstruction methods, the HLL solver filled ``reconstruction::InterfaceState::scalar[i]`` with a scalar density. Furthermore, the associated flux calculations also assumed that the quantity is a scalar density.

This means that with piecewise constant reconstruction, the HLL solver would return an incorrect scalar flux.

### DIFF
--- a/src/riemann_solvers/hll_cuda.cu
+++ b/src/riemann_solvers/hll_cuda.cu
@@ -71,7 +71,7 @@ __global__ void Calculate_HLL_Fluxes_CUDA(Real const *dev_conserved, Real const 
       left_state.energy       = dev_bounds_L[4 * n_cells + tid];
 #ifdef SCALAR
       for (int i = 0; i < NSCALARS; i++) {
-        left_state.scalar[i] = dev_bounds_L[(5 + i) * n_cells + tid];
+        left_state.scalar[i] = dev_bounds_L[(5 + i) * n_cells + tid] / left_state.density;
       }
 #endif
 #ifdef DE
@@ -85,7 +85,7 @@ __global__ void Calculate_HLL_Fluxes_CUDA(Real const *dev_conserved, Real const 
       right_state.energy       = dev_bounds_R[4 * n_cells + tid];
 #ifdef SCALAR
       for (int i = 0; i < NSCALARS; i++) {
-        right_state.scalar[i] = dev_bounds_R[(5 + i) * n_cells + tid];
+        right_state.scalar[i] = dev_bounds_R[(5 + i) * n_cells + tid] / right_state.density;
       }
 #endif
 #ifdef DE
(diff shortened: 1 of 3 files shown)


evaneschneider:
Thanks for catching this!

PR #533, opened 2026-08-17
Metallicity Dependent Cooling and metal_density scalar field

hannahjleary:
### Summary
Adds metallicity-dependent cooling and a new scalar `metal_density` field to track metal content per cell.

### Changes
-Added `metal_density` as a new tracked scalar field (using `SCALAR` and `METALS` macros). This is used by `CoolRecipeMetals` to compute the metallicity of a cell as `metal_mass_frac / solar_metal_mass_frac`.
-`CoolRecipeMetals`: Combines primordial cooling with metal-line cooling by scaling the metal-line contribution of a solar abundance Cloudy model by the cell's metallicity.
-`CoolRecipePrimordial`: added a recipe to use `cool_component::primordial_cool`. Requires heating to be disabled by setting a new parameter `chemistry.enable_heating` to `false` (default: `true`)
-All cooling recipes now take a hydrogen and helium number density (`n_H` and `n_He`, respectively), as well as a metallicity.
-Added documentation for the two new cooling recipe options, the `METALS` macro, and the `metal_density` output fields
  -`docs/sphinx/Physics/metallicities.png` shows the cooling rates of `CoolRecipeMetals` for a range of metallicities.
  -`docs/sphinx/Physics/models.png` shows the primordial cooling curve, the `CoolRecipeMetals` curve, and an actual Cloudy model for four different metallicities. Our model closely follows the true Cloudy run even in the low and high metallicity limits.
-The initial conditions function `Clouds()` now takes two new input parameters from `pmap`: `metallicity_wind` and `metallicity_cloud` (default value: `1.0`)
-The initial conditions function `Constant()` now takes one new input parameter from `pmap`: `metallicity` (default value: `1.0`)


config/make.type.cosmology
```diff
@@ -24,9 +24,6 @@ DFLAGS += -DCOSMOLOGY
 # Average Slow cell when the cell delta_t is very small
 DFLAGS += -DAVERAGE_SLOW_CELLS
 
-# Solve the Gas Internal Energy using a Dual Energy Formalism
-DFLAGS    += -DDE
-
```
mabruzzo, line 29: Was this an intentional change? Or something that you just tweaked while debugging?

config/setup.crc.gcc.sh
```diff
@@ -3,7 +3,8 @@
 #-- This script needs to be sourced in the terminal, e.g.
 #   source ./setup.crc.gcc.sh
 
-module load python/anaconda3-2020.11 gcc/10.1.0 cuda/11.1.0 openmpi/4.0.5 hdf5/1.12.0 googletest/1.11.0
+# module load python/anaconda3-2020.11 gcc/10.1.0 cuda/11.1.0 openmpi/4.0.5 hdf5/1.12.0 googletest/1.11.0
+module load cuda/11.8 gcc/11.5.0-dak7qob openmpi/4.1.7-cuda hdf5/1.14.6 # for a100 nodes
```
mabruzzo, line 7: we should probably revert this. Maybe the better alternative would be to create a new setup file explicitly dedicated to the a100 nodes?

config/make.type.hydro
```diff
@@ -27,6 +34,7 @@ DFLAGS    += -DTEMPERATURE_FLOOR
 # Can also add -DSLICES and -DPROJECTION
 OUTPUT    ?=  -DOUTPUT -DHDF5
 DFLAGS    += $(OUTPUT)
+DFLAGS    += -DSLICES -DPROJECTION
```
mabruzzo, line 37: we should either revert this change or get rid of the comment `# Can also add -DSLICES and -DPROJECTION`

config/make.type.hydro
```diff
@@ -10,15 +10,22 @@ DFLAGS    += -DSIMPLE
 DFLAGS    += -DPPMP
 else
 DFLAGS    += -DVL
-DFLAGS    += -DPLMC
+DFLAGS    += -DPPMP
```
mabruzzo, line 13: I think we should revert this.

```suggestion
DFLAGS    += -DPLMC
```

Out of curiosity, why are you using PPMP? Why not PLMP?

config/make.type.hydro
```diff
@@ -10,15 +10,22 @@ DFLAGS    += -DSIMPLE
...
 # Solve the Gas Internal Energy using a Dual Energy Formalism
-#DFLAGS    += -DDE
+DFLAGS    += -DDE
+
+
+# Define metals macro and evolve additional scalars
+DFLAGS    += -DSCALAR
+DFLAGS    += -DSCALAR_FLOOR
+DFLAGS    += -DMETALS
+#DFLAGS    += -DBASIC_SCALAR
```
mabruzzo, line 28: I think we should revert this file (to avoid breaking the test-answers). Perhaps the answer is to make a new configuration file.

src/global/global.h
```diff
@@ -95,7 +95,11 @@ typedef double Real;
 #define INITIAL_FRACTION_HEII     9.59999999903e-15
 #define INITIAL_FRACTION_HEIII    9.59999999903e-18
 #define INITIAL_FRACTION_ELECTRON 1.53965115054e-4
-#define INITIAL_FRACTION_METAL    1.00000000000e-10
+#define INITIAL_FRACTION_METAL    1.00000000000e-10   
+
+// Solar Metal Mass Fraction
+#define SOLAR_METAL_MASS_FRAC     0.01295
+#define HYDROGEN_FRAC_BY_MASS     0.716
```
mabruzzo, line 102: This constant is now slightly inconsistent with Brant's newly added `"YHe"` runtime parameter (the Helium-fraction-by-mass). It's worth mentioning that his parameter is currently cosmology-only.

The way I see it, we have a few options:
1. we kick the can down the road and don't deal with this conflict right now. If we do that then we need to add some comments here where we declare `HYDROGEN_FRAC_BY_MASS` and make it clear that this is only temporary, incompatible with the `"YHe"` runtime parameter, and will need to be replaced at some point in the future
2. We can try to adjust things now to make them consistent. (A way to handle this may involve defining a "Model" dedicated to tracking these sorts of physical constants)

One point I will make, is that if we want to change how the `"YHe"` parameter works (perhaps replacing it with hydrogen fraction by mass), we probably we want to make that change sooner rather than later (while reviewing Brant's PR, I highlighted that this could need to change). If we instead decide we want to the behavior in the future (say in a few years), that could be quite annoying for people running cosmological simulations.

TO BE CLEAR: it's totally fine with me if we put off dealing with this inconsistency, as long as we add comments to document that the inconsistency exists.

src/grid/boundary_conditions.cpp
```diff
@@ -522,8 +522,9 @@ void Grid3D::Custom_Boundary(char bcnd[MAXLEN])
 
 /*! \fn void Wind_Boundary()
  *  \brief Apply wind boundary */
-void Grid3D::Wind_Boundary()
+void Grid3D::Wind_Boundary(ParameterMap &pmap)
 {
+  Real metallicity_wind = pmap.value_or("metallicity_wind", 1.0);
   int x_off, y_off, z_off;
```
mabruzzo, line 528: I really don't love the way that we are passing `ParameterMap` as an argument to all of these boundary handling functions. 

- to elaborate on this, `ParameterMap` is intended to be a short term object that only exists while we start up the simulation. It has machinery built into (that isn't presently used)[^1] it in order to check whether parameters have been read from it. This only works if we know we are completely done using the `ParameterMap` and don't need to access it again. Unfortunately, if we need to pass `pmap` to all of the boundary-enforcement methods, that means that `ParameterMap` must persist for the entirety of the simulation
- As a short term solution, the easiest way to handle this is to simply store the value of the `"metallicity_wind"` inside of `Parameters`.
- The longer term solution is to track the value of `"metallicity_wind"` inside of a cloud model type and access it from there (in fact this was one of the scenarios I was explicitly considering when I designed the model types).

My preference would be for you to stop passing `ParameterMap` to all of the boundary-related functions in this PR. I realize that this involves a fair amount of work. If you want to put this off until after you make the `CloudModel` class, that's probably ok. But if somebody else comes along and starts writing code depending on the fact that you pass `ParameterMap` to all of these functions between now and then, you're going to need deal with refactoring all that code.


[^1]: I dug into this. The only reason this isn't currently used was because of that period where there were 2 different ways to access parameters. When we transitioned to only having 1 approach, I totally forgot to start enforcing the check that all parameters in `ParameterMap` are actually used. I'll add fixing this to my todo list

src/cooling/cooling_cuda.cu
```diff
@@ -154,28 +156,48 @@ __global__ void cooling_kernel(Real *dev_conserved, int nx, int ny, int nz, int
...
-    cool = recipe.cool_rate(n, T);
+    cool = recipe.cool_rate(n_H, n_He, T, metallicity);
 
     // calculate change in temperature given dt
     del_T = cool * dt * TIME_UNIT * (gamma - 1.0) / (n * KB);
 
+    if (xid == is && yid == js && zid == ks) {
+      // printf("cool = %e, del_T = %e\n", cool, del_T);
+    }
+
```
mabruzzo, line 200:
```suggestion
```

src/cooling/cool_components.h
```diff
@@ -52,9 +52,10 @@ namespace cool_component
 {
 
 /*! Primordial hydrogen/helium cooling curve (derived according to Katz et al. 1996.) */
-inline __device__ Real primordial_cool(Real n, Real T)
+inline __device__ Real primordial_cool(Real n_H, Real n_He, Real T)
 {
-  Real n_h, Y, y, g_ff, cool;
+  // Real n_h, Y, y, g_ff, cool;
```
mabruzzo, line 57:
```suggestion
```

src/cooling/cooling_cuda.cu
```diff
@@ -82,20 +83,20 @@ class CoolingUpdateExecutor
...
   }
 };
 
 /*! \fn void cooling_kernel(Real *dev_conserved, int nx, int ny, int nz, int
- n_ghost, int n_fields, Real dt, Real gamma, cudaTextureObject_t coolTexObj,
+ n_ghost, int n_fields, Real dt, Real gamma, Real solar_metal_mass_frac, Real hydrogen_frac_by_mass, cudaTextureObject_t coolTexObj,
  cudaTextureObject_t heatTexObj)
  *  \brief When passed an array of conserved variables and a timestep, adjust
  the value of the total energy for each cell according to the specified cooling
  function. */
```
mabruzzo, line 96: This seems like a great opportunity to simplify this comment block to stop repeating the function signature. This is totally unnecessary. In fact I think we should eliminate this repetition wherever possible because the repeated function signature tends to get out-of-sync with the actual signature.

For the sake of clarity, after these changes the docstring should look like:

```c++
/*! \brief When passed an array of conserved variables and a timestep, adjust
 the value of the total energy for each cell according to the specified cooling
 function. */
```

mabruzzo (commented):
This looks fantastic!

There are a couple of changes I would like to see. And then I would like to take another look. But you are VERY close!

PR #534, opened 2026-08-21
Reconcile cosmo-ics with dev

brantr:
The merge process with dev broke a few things with cosmo-ics, which this PR attempts to fix. This is a small, targeted PR meant to actually enable the cosmological ICs to run.

* The behavior of the Runge-Kutta scheme changed. We've reinitialized a variable inside a loop and it seems to work.
* A check in `chemistry_functions.cpp` now is triggered regarding how / whether `H_fraction` is set. I think there is an initialization step that must have been removed somehow. I've added this back.
* There were two `YHe` declared in global.h.

Minor item:
* Renamed `cosmoics_seed` to `cosmo_ics_seed`.

Output directory
* The Parameter member `P.outdir` was removed and replaced with `writer_manager.fname_template().nominal_output_dir_path().c_str()`. This is not very user friendly, and seems to require the API to the initialization functions to be re-written if you want to write a file during initialization. Instead of making those more extensive changes, I added `P.outdir` back and then just copy the writer manager result into it. I don't seem the harm in this, so hopefully that is a minor addition that will pass muster.

### DIFF
--- a/config/make.type.cosmology
+++ b/config/make.type.cosmology
@@ -10,11 +10,11 @@ DFLAGS += -DCOSMOLOGY
 
 
 # Include the FFT subroutines
-#DFLAGS += -DFFT
+DFLAGS += -DFFT
 
 
 # Solve the Primordial Chemical Network (H+He) on the GPU (Includes Radiative Cooling, Photoheating and Photoionization)
-#DFLAGS += -DCHEMISTRY_GPU -DOUTPUT_TEMPERATURE -DOUTPUT_CHEMISTRY
+DFLAGS += -DCHEMISTRY_GPU -DOUTPUT_TEMPERATURE -DOUTPUT_CHEMISTRY
 
 
 # Perform In-The-Fly analysis of Cosmological Simulations
--- a/config/make.type.cosmology
+++ b/config/make.type.cosmology
@@ -10,11 +10,11 @@ DFLAGS += -DCOSMOLOGY
 
 
 # Include the FFT subroutines
-DFLAGS += -DFFT
+#DFLAGS += -DFFT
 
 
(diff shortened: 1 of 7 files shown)


brantr:
pre-commit.ci autofix

mabruzzo:
FYI: This is on my radar.

I haven't reviewed it yet because I haven't fully thought through the best way for us to handle:

> The Parameter member `P.outdir` was removed and replaced with `writer_manager.fname_template().nominal_output_dir_path().c_str()`. This is not very user friendly, and seems to require the API to the initialization functions to be re-written if you want to write a file during initialization. Instead of making those more extensive changes, I added `P.outdir` back and then just copy the writer manager result into it. I don't seem the harm in this, so hopefully that is a minor addition that will pass muster.

You are absolutely right that `writer_manager.fname_template().nominal_output_dir_path().c_str()` is not very user friendly. At the time that I refactored things, I suspect my thought process was that writing files at initialization was an unusual/atypical thing for us to do. Clearly, I was wrong.

I really don't want to track `outdir` in `Parameters` struct.[^1] Instead, I *think* the better solution is to add a method to `WriterManager` so we could write `writer_manager.root_output_dir()`.[^2]  But, I'm going to take another another day or so to mull things over.

In the particular instance where we are printing out the output directory during startup, I have an idea for more thoroughly printing out the simulation's runtime-configuration (but that's a topic for another time)

[^1]: As I've stated before, I think we should dispose of the `Parameters` struct. This is admittedly a somewhat slow and gradual process as we figure out the optimal home for certain quantities that must be accessed in various corners of the codebase (this was part of the motivation for introducing the model machinery).
[^2]: Perhaps this is a sign that `WriterManager` should actually be called `OutputManager`.

PR #535, opened 2026-08-24
A few tweaks to satisfy python linter

mabruzzo:
This PR makes a few very minor tweaks to satisfy the  python-linter. I suspect that this was causing pre-commit.ci to fail


mageejake (approved):
Fantastic PR, Matthew!

PR #536, opened 2026-08-24
Fix GHA containers (for real this time)

mabruzzo:
To be reviewed after #535 is merged

------

This PR attempts to adjust the GHA containers to correctly be able to run clang-tidy. We are getting awfully close (and I've done a bunch of local testing that suggests that everything will work).

PR #537, opened 2026-08-25
Run clang-tidy within GitHub Actions

mabruzzo:
Before this PR can be reviewed,  we must:
- review and merge in PR #536
- after #536 is merged, we need to trigger a new build of docker images off of the dev branch
- a commit must be pushed to this branch to reference the new docker image versions

Before this can be reviewed, we also need to merge in #529 and #535 (the changes from those PR have been included as part of the current PR in order to address potential modes of CI-failure)

-----------

This PR takes massive steps to run clang-tidy within GitHub Actions. I've tested everything locally and it looks like it will all work (there's a chance I may have missed a configuration mode, but I'm confident that I've checked all of them).

The changes in this PR:
- we now create a `build` directory where we can stick all artifacts generated as part of a build. Currently, this is just used to hold `compile_commands.json` (see next bullet-point), but will be useful in other contexts.
- we now create a [`compile_commands.json` "compilation database"](https://clang.llvm.org/docs/JSONCompilationDatabase.html) with a new script called `tools/generate_compile_commands.py`.
  - this is generally used for telling development tools (especially tools provided by LLVM) how the flags that are required to compile a given source file
  - FYI: with a more-modern build-system (e.g. CMake or Meson), this is created automatically.
  - this file is placed inside of the `build` subdirectory. It *should* be regenerated every time that any part of the build-configuration changes, but (like most other parts of our build) that's not strictly enforced
- the `build/compile_commands.json` file is now our primary approach for telling `clang-tidy` about the compilation flags (rather than passing it all the flags for every invocation). This has a few benefits:
  1. `clang-tidy` can be invoked one time for all of the targets (rather than once for GPU-source files and once for C++ source files). This has a few consequences for the `tidy` makefile-target
     -  Historically, we needed a whole show and dance to run properly run things, which was a little bug-prone (as PR #526 illustrated).
     - Relatedly, the `tidy` makefile-target can now directly show the results of clang-tidy (previously we had to do weird contortions in CI to show the output).
     - For historical consistency, we still write the output of the `tidy` makefile-target to an output file. But, I'm tempted to stop doing that (if someone really wants to save the output to disk, they can just pipe the output to disk themselves).
  2. it is now possible to invoke `clang-tidy` without using the `tidy` makefile-target.
- I have removed the logic for invoking `clang-tidy` on Jenkins and added logic to the compilation-check workflow on GHA
  - to make this workflow as fast as possible, we actually invoke the `run-clang-tidy.py` script that is shipped alongside `clang-tidy`.
    - This script accepts distributes the actual work of invoking clang-tidy between multiple `clang-tidy` instances (in contrast, the current `tidy` makefile-target only uses one clang-tidy instance at a time). The `-j` flag in specifies the number of `clang-tidy` instance to use at any given times.
    - We don't use `run-clang-tidy.py` inside the Makefile for 2 reasons because way that parallelism is handled by GNU Make is incompatible with `run-clang-tidy.py`'s j-flag
    - Our use of `run-clang-tidy.py` is ONLY possible outside of the Makefile because of our use of `compile_commands.json`\
  - we do not currently run clang-tidy with the HIP compiler. It's probably something that we want to do, but that definitely involves a few tweaks -- essentially, we need to transform the flags that would be passed to `hipcc` in a similar (but different) way that we transform flags that would be passed to `nvcc`
- as a convenience, I added the `PHONY` `setup`-target to the Makefile to make it easier to trigger the creation of `compile_commands.json`. This is primarily intended to help invoke `run-clang-tidy.py` (I'm on the fence about whether we rename this target).

>[!note]
> All GHA checks involving clang-tidy will fail for as long as the prerequisites pertaining to #536 remain addressed (until that happens, these checks all point to images where clang-tidy is broken).

In the background, I'm plugging away at a few not strictly-essential improvements (e.g. making the build-system a little more robust, better documenting how exactly the compile documentation for how to run, making modifications to the build-system a little more robust). But, this should be reviewed and merged ASAP (if those improvements aren't ready, they will be shifted to a separate PR).


helenarichie (approved):
This looks great to me! And I think the `setup` target name is okay as-is

